PR #2, opened 2019-01-20
Add 'zupo' quote

kernc:
Source: https://twitter.com/nzupan/status/1086246420586336257


kernc:
Thanks @offlinehacker. I don't think @zupo will mind at all. :+1: :heart:

PR #7, opened 2019-01-23
WiP: Add support for PEP 484 type hints

### LINKED ISSUE #6: Support PEP 484 type hints
### Expected Behavior

When generating doc for a package that is annotated with [PEP 484] type hints, I would expect the type hints to be present in the documentation.

[PEP 484]: https://www.python.org/dev/peps/pep-0484/

For example, the following definition
```python3
def foo(x: Bar, repeat : int = 1):
    """Foo-cate a Bar a number of times (default: 1)."""
    ...
```

results in the type hints being available in the REPL's `help()`:
```python3
>>> help(foo)
foo(x: __main__.Bar, repeat: int = 1)
    Foo-cate a Bar a number of times (default: 1).
```

and it should well as be present in the pdoc-generated documents:
```markdown
$ pdoc example.py
Module example
==============

Functions
---------

`foo(x: Bar, repeat: int = 1)`
:   Foo-cate a Bar a number of times (default: 1).

Classes
-------

`Bar`
```

### Actual Behavior

`pdoc` ignores the type-hints:
```markdown
$ pdoc example.py
Module example
==============

Functions
---------

`foo(x, repeat=1)`
:   Foo-cate a Bar a number of times (default: 1).

Classes
-------

`Bar`
```

Type hints document the expected types of parameters and return values.  As such, they convey a lot of information for the documentation's user, when present.

See [PEP 484] for a discussion of type hints.

### Steps to Reproduce

Generate documentation for a module that uses type hints

### Additional info

- pdoc version: 0.5.1
- Python version: 3.7.2 (default, Jan  3 2019, 02:55:40)

nbraud:
Fixes #6.

I started by refactoring `Function.params` to produce a list of `Parameter` objects.
This was necessary, as type hints aren't represented the same way in text and HTML (due to linking).

Feel free to comment on the refactor, but type hints themselves are very much a work in progress (the pushed version just adds the necessary field to `Parameter`, that's all).

- [x] Refactor `Function.params`.
- [x] Render type hints for function parameters (in plain text);
- [x] Render type hints for function return values.
- [ ] Render type hints for instance variables & properties.


nbraud:
Ah, I used language features only available with 3.6 and later.  Is that an issue?

kernc:
At first look, I find this far too abstracted for what is gained or what pdoc is meant to accomplish. While this sure looks _correct_, I'd honestly _much_ prefer simpler:
```py
def annotated_params(self) -> List[Tuple[str, str, str]]
```
with, by wrapping:
```py
def foo(_a, b: str, *args, c: int = None, _d=2, **kwargs): pass
```
returning:
```py
[('_a',       None,  None),
 ('b',        'str', None),
 ('*args',    None,  None),
 ('c',        'int', 'None'),
 ('**kwargs', None,  None)]
```
Maybe as simple namedtuples. Then you can extend `Function.params(include_types=False)` and the whole thing remains backwards-compatible with templates needing just pass a simple config option to `params()` call. Some type hints can be very complexly nested and meant primarily for linters to verify, so inclusion should certainly be optional up to each project.

Also, Python 3.5, so no variable type annotations, dataclasses, or f-strings.

nbraud:
> def annotated_params(self) -> List[Tuple[str, str, str]]

The `Union` was something I forgot to remove while working on the PR, it was meant to just be `List[Parameter]`.  Anyway, making `Parameter` a NamedTuple, as requested.

b52:
How is this coming along? I really would like to see this feature. Do you need help or anything @nbraud ?

pdoc/__init__.py
```diff
@@ -636,11 +638,16 @@ def _var_docstrings(doc_obj: Union['Module', 'Class'], *,
     return vs
 
 
-def _is_public(ident_name):
+def _is_public(ident):
     """
     Returns `True` if `ident_name` matches the export criteria for an
```
b52: Should be `ident`

nbraud:
@b52 I had some medical stuff happen and was entirely-useless this week.

For what it's worth, I started rewriting this PR to use [`inspect.signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) instead:
- it seems like a much more convenient way to get to the necessary (meta)data;
- it addresses @kernc's concerns about abstraction: we just get the right abstractions from the standard library instead of rolling our own;
- [`getfullargspec`](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec), which is currently used, was deprecated in favour of it.

kernc:
@nbraud Hate to sound insensitive to your personal issues, but love for you to continue this. :sun_with_face:

nbraud:
@kernc It's fine, I appreciate it's been a while.  I finally had the time & energy to look into it again, though it's not yet complete as I broke some tests in the process.

The good news is that `Function.params` got much less complicated  :)

nbraud:
@kernc OK, I think I got minimum, viable PEP484 support going, but a seemingly-unrelated test fails:
```
======================================================================
ERROR: test_http (pdoc.test.HttpTest) (url='/csv.ext')
----------------------------------------------------------------------
  File "/home/nicoo/devel/py/ppb/pdoc/pdoc/test/__init__.py", line 891, in test_http
    with urlopen(url + 'csv.ext', timeout=3) as resp:
  File "/usr/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.7/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib/python3.7/urllib/request.py", line 641, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib/python3.7/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.7/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```

pdoc/__init__.py
```diff
@@ -1366,62 +1414,30 @@ def _is_async(self):
             return False
 
     @lru_cache()
-    def params(self) -> List[str]:
+    def params(self) -> List[Union[str, Parameter]]:
```
kernc: I'm strongly opposed to breaking established API by returning something else than strings. It makes no sense, really, as any code that were to do something more than `str(param)` will have to instance check.

Maybe is better:
```py
def params(self, *, types: bool = False) -> List[str]
```
with `types=True` returning the strings interpolated with typing info?

pdoc API-wise, I think it's easy enough to let the user do on their own if needed:
```python
inspect.signature(pdoc_func_obj.obj)
```
In that way, `class Parameter(inspect.Parameter)` is not even required, in exchange for some of the former complexity of this method (should be much less with `inspect` API).

kernc: `types=` parameter can then be a template tunable in _config.mako_:
https://github.com/pdoc3/pdoc/blob/a1f0a15c46e7698c1c92607d650d4487c9143c67/pdoc/templates/config.mako#L1-L12

kernc:
@nbraud Do you agree with my comments? I'd love to see this brought to completion.

kernc:
Replaced with https://github.com/pdoc3/pdoc/commit/5df06cde92538269bf53b3a6072e9f0aff0b3235. Thanks anyway, @nbraud. I'd appreciate feedback whether the result is adequate.

Issue #4, opened 2019-01-22
bash: pdoc: command not found


wq2012:
### Expected Behavior

On a Mac or Linux machine,
```
$ pip3 install pdoc3
$ pdoc --html pdoc
```
produces documentation. 

### Actual Behavior

Second command yields:
```
$ bash: pdoc: command not found
```

### Steps to Reproduce

1. On a Linux or Mac machine, open terminal
2. Type the commands
3. Read the results

### Additional info

- pdoc version:
```
Python 3.5.4rc1 (default, Jul 25 2017, 08:53:34)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pdoc
>>> print(pdoc.__version__)
0.5.1
```

### Suggestion
Add ` __main__.py` file such that users can at least do:

```
$ python3 -m pdoc ...
```

kernc:
I'm sorry, I can't reproduce the issue:
```bash
$ pip3 install pdoc3
Collecting pdoc3
...
Successfully installed MarkupSafe-1.1.0 mako-1.0.7 markdown-3.0.1 pdoc3-0.5.1

$ which pip3
/usr/bin/pip3

$ pip3 freeze | grep pdoc
pdoc3==0.5.1

$ which pdoc
/home/user/.local/bin/pdoc

$ which pdoc3
/home/user/.local/bin/pdoc3

$ cat $(which pdoc)
```
```py
#!/usr/bin/python3

# -*- coding: utf-8 -*-
import re
import sys

from pdoc.cli import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
As setup.py _does_ specify [console scripts](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point), it should be up to setuptools/pip to wire up the executables. 
https://github.com/pdoc3/pdoc/blob/e7868e286c260650daeee478a8b2f1e9ed4c8767/setup.py#L39-L44
Have you tried updating pip and setuptools? Do you have any other ideas?

> Add `__main__.py` file such that users can at least do

Would you care to add `__main__.py` in a pull request?

wq2012:
Updating pip3 didn't seem to resolve the issue for me. Not sure why.

Anyway, adding __main__.py should make it easier. I've sent pull request. Hope we can include it in the next PyPI release.

[...]

PR #8, opened 2019-01-23
[Issue #4] Add __main__.py such that users can do python -m pdoc

wq2012:
Fixes https://github.com/pdoc3/pdoc/issues/4.

With this changes, users can do:

```
$ python3 -m pdoc
usage: __main__.py [-h] [--version] [--filter STRING] [--html]
                   [--html-dir DIR] [--html-no-source] [--overwrite]
                   [--external-links] [--template-dir DIR]
                   [--link-prefix STRING] [--http HOST:PORT]
                   MODULE [MODULE ...]
```

Thus even if the setuptools failed to generate the executable console binary (which happened to me on Mac and Linux), the users are still able to use the tool, without having to clone and touch the code.

### DIFF
--- a/pdoc/__main__.py
+++ b/pdoc/__main__.py
@@ -1,0 +1,6 @@
+#!/usr/bin/env python3
+
+from pdoc import cli
+
+if __name__ == "__main__":
+    cli.main(cli.parser.parse_args())
--- a/pdoc/__main__.py
+++ b/pdoc/__main__.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from pdoc import cli
+from pdoc.cli import main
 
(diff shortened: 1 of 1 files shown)


pdoc/__main__.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python3
+
+from pdoc import cli
+
+if __name__ == "__main__":
+    cli.main(cli.parser.parse_args())
```
kernc: This is nitpicking, but would you mind just importing `main` and calling it without arguments? `cli.parser`, obviously, is an implementation detail.

wq2012: Done.

kernc:
Excellent, thank you!

PR #13, opened 2019-01-25
Make source code button less dominant

b52:
Personally, I find a source code button presented this way very distracting without providing any meaningful benefit. Thus I tried to make it less prominent in the rendered HTML.
Additionally, I removed a superfluous bracket that was rendered into the code view.

Here are some screens how it would look with this PR:
image
image


kernc:
This is much better indeed! Merged with only a slight modification. Thanks!

PR #16, opened 2019-01-31
ENH: Support for include reST directive

### LINKED ISSUE #9: Feature suggestion: Getting index.html contents from a markdown file
### Expected Behavior

Well, it's not something important, but a nice-to-have feature.

Currently the `<article>` part of `index.html` comes of the docstring of `__init__.py`. However, some authors prefer to leave the details of the package/module in a markdown file like `README.md`, for easier access to the audience, instead of `__init__.py`.

It might be a nice feature if `pdoc` has an argument that supports converting a markdown file to HTML and injecting the converted HTML part into the `<article>` part of `index.html`.

### Actual Behavior

N/A

### Steps to Reproduce

N/A

### Additional info

- pdoc version:
0.5.1

b52:
Fixes #9


pdoc/html_helpers.py
```diff
@@ -227,20 +233,31 @@ def _admonition(match):
...
+                                        r'((?:\n(?:(?P=indent) +.*| *$))*)', re.MULTILINE).sub,
+                             partial(_ToMarkdown._admonition, module))
         # Apply twice for nested (e.g. image inside warning)
-        return _admonitions(_admonitions(text))
+        return substitute(substitute(text))
+
+    @staticmethod
+    def include(path: Path, indent: str) -> str:
+        if not path.exists():
+            raise RuntimeError('include %s: file does not exist' % (path,))
```
kernc: I'd prefer just:
```py
try:
    with open(path) as f:
        return ''.join(indent + line for line in f)
except Exception as e:
    raise RuntimeError('`.. include:: {}` error in module {!r}: {}'.format(value, module, e))
```
Since Python 3.4, I believe, this will print both exceptions, ours, detailed, last.

pdoc/html_helpers.py
```diff
@@ -227,20 +233,31 @@ def _admonition(match):
...
+
+        if path.suffix not in ('.md', '.markdown'):
+            raise RuntimeError('include %s: only markdown sources allowed '
```
kernc: :-1: this limitation. Could be .txt and still formatted as markdown. Better just document verbatim inclusion.

pdoc/__init__.py
```diff
@@ -185,8 +185,8 @@ def __init__(self):
 Currently, pure Markdown (with [extensions]), [numpydoc],
 and [Google-style] docstrings formats are supported.
 Basic reST directives (such as e.g.
-`.. versionadded::`, `.. deprecated::`, `.. note::`, `.. image::`, ...)
-should also work.
+`.. versionadded::`, `.. deprecated::`, `.. note::`, `.. image::`,
+`.. include::`, ...) should also work.
```
kernc: Welcome to extend on it with a dedicated subsection:
```md
### Supported reST directives
```
containing a list, mentioning any particularities.

pdoc/html_helpers.py
```diff
@@ -198,10 +199,14 @@ def google(text,
         return _googledoc_sections(text)
 
     @staticmethod
-    def _admonition(match):
+    def _admonition(module, match):
         indent, type, value, text = match.groups()
 
-        if type in ('image', 'figure'):
+        if type == 'include' and module:
+            options = dict(re.findall(r'\n *:([^:]+): *([^\n]+)', text, re.MULTILINE))
```
kernc: Parse options from within `include()` with a dedicated function. Maybe can be reused for others (`image` etc.).

pdoc/html_helpers.py
```diff
@@ -227,20 +232,34 @@ def _admonition(match):
...
+    def include(path: Path, indent: str, options: dict) -> str:
+        if not path.exists():
+            raise RuntimeError('include %s: file does not exist' % (path,))
+
+        if path.suffix not in ('.md', '.markdown'):
+            raise RuntimeError('include %s: only markdown sources allowed '
+                               'for now' % (path,))
+
+        selection = slice(int(options.get('start-line', 0)),
+                          int(options.get('end-line', 0)) or None)
```
kernc: I'm an awful opponent of manual managing. It's bound to get out of sync. Can we instead support [`start-after` and `end-before`](http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment) with some boundary [markdown or HTML comments](https://stackoverflow.com/questions/4823468/comments-in-markdown)? Then third party documents won't need to get updated every time a line of prose overflows.

pdoc/html_helpers.py
```diff
@@ -198,10 +199,14 @@ def google(text,
...
 
     @staticmethod
-    def _admonition(match):
+    def _admonition(module, match):
         indent, type, value, text = match.groups()
 
-        if type in ('image', 'figure'):
+        if type == 'include' and module:
+            options = dict(re.findall(r'\n *:([^:]+): *([^\n]+)', text, re.MULTILINE))
+            path = Path(module.module.name.replace('.', '/')) / value
```
kernc: See the outdated comments. Certainly no pathlib. Prefer `os.path.join(module.obj.__file__, value)`, but within `include()`.

pdoc/__init__.py
```diff
@@ -185,8 +185,9 @@ def __init__(self):
 Currently, pure Markdown (with [extensions]), [numpydoc],
 and [Google-style] docstrings formats are supported.
 Basic reST directives (such as e.g.
-`.. versionadded::`, `.. deprecated::`, `.. note::`, `.. image::`, ...)
-should also work.
+`.. versionadded::`, `.. deprecated::`, `.. note::`, `.. image::`, ...) should
+also work, in addition to `.. include::` with its `:start-line:` and
+`:end-line:` options.
```
kernc: With options support, I request aforementioned subsection. :smile:

pdoc/html_helpers.py
kernc: By that I mean I opinionatedly _refuse_ to support something as change-prone as line numbers. Strip. :blush:

kernc:
> query the module list

Do you mean when running in dev HTTP web server mode (`--http`)?

pdoc/html_helpers.py
kernc: Could also be used to include examples, e.g.:
~~~
```py
.. include:: some_example.py
```
~~~

pdoc/html_helpers.py
```diff
@@ -198,10 +199,15 @@ def google(text,
         return _googledoc_sections(text)
 
     @staticmethod
-    def _admonition(match):
+    def _admonition(module, match):
         indent, type, value, text = match.groups()
 
-        if type in ('image', 'figure'):
+        if type == 'include':
+            path = Path(value)
```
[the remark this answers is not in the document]
b52: Just out of curiosity, could you elaborate why?

pdoc/html_helpers.py
b52: I guess it depends on the use case. For me I just want to drop the first line (header) of the `README.md` that I include. I think it's highly unlikely that my `:start-line: 1` gets out of sync. And with everything, it's up to the user to make sure the documentation is accurate, as well as its options.

b52: I implemented both options now. I'm highly in favor of `start-line`, because I absolutely do not want to augment my `README.md` and I'm very certain that my h1 is always on the first line ;)

pdoc/html_helpers.py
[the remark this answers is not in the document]
kernc: It's just not used anywhere else in the project, and for this one line ...

pdoc/html_helpers.py
```diff
@@ -227,20 +230,60 @@ def _admonition(match):
...
+
+        end_before = options.get('end-before')
+        if end_before:
+            for end in range(end, start - 1, -1):
+                j = lines[end].find(end_before)
+                if j >= 0:
+                    lines[end] = lines[end][:j]
+                    break
+
+        return ''.join(lines[start:end + 1])
```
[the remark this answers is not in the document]
b52: Not exactly, cause you have to handle the case when the pattern is not contained. But to be honest, I do not really know what to do in this case anyways. Currently it just returns an empty string, if the `start_after` is not found.

pdoc/html_helpers.py
```diff
@@ -227,20 +230,60 @@ def _admonition(match):
...
+                lines = [indent + line for line in f]
+        except Exception as e:
+            raise RuntimeError('`.. include:: {}` error in module {!r}: {}'
+                               .format(value, module, e))
+
+        options = _ToMarkdown._parse_directive_options(text)
+
+        start = int(options.get('start-line', 0))
+        try:
+            end = int(options['end-line'])
```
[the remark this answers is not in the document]
b52: Yeah, it was late :D

b52:
@kernc Could you maybe chime in here and fix the test? I tried to add one but that results in a lot of not found errors.

pdoc/html_helpers.py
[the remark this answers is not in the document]
kernc: Ah, I didn't bother to confirm. `str.find()` indeed returns `-1`. We should instead use `str.index()` and `str.rindex()`. Then, the implementation will _loudly fail_ when an option is specified but can't be honored. As it should.

kernc:
I did chime in. Please verify.

The reason the tests were failing is that the test module layout is a handful of symlinks to one single module. So your `reST_directives` test with `.. include::` was looking for relative `subpkg/include_me.py` from all sorts of directories.

It seemed the easiest to create a separate test.

b52:
@kernc Is there still anything blocking this?

kernc:
Nothing at all. It's great. I didn't think we would even get options in that easily. Thanks!

PR #25, opened 2019-02-07
Makes closing stdin optional and false by default

### LINKED ISSUE #22: Make closing stdin optional?
For the module I'm working on, this defensive line is actually breaking things because a dependency of a dependency is checking if `stdin` is open: https://github.com/pdoc3/pdoc/blob/master/pdoc/cli.py#L285

Would it be possible to allow this behaviour to be turned off via a flag or environment variable? If you'd accept such a PR I'd be happy to contribute it :)

nicolaskruchten:
Closes #22

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -80,6 +80,12 @@ aa(
          "No link prefix results in all links being relative. "
          "No effect when combined with --http.",
 )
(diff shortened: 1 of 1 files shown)


kernc:
Thanks.

PR #26, opened 2019-02-08
ENH: optional sorting of variables, classes, functions

### LINKED ISSUE #23: Make identifiers sorting optional
Would you accept a PR to make sorting optional in the documentation (new CLI arg, defaulting to current behaviour)? I rather like the order in which the functions/methods are declared in my module and I'd rather the documentation follow that order rather than a strictly alphabetical one :)

nicolaskruchten:
Closes #23 

I wasn't able to make module-sorting optional as it seems to happen while walking the file tree, and I didn't really know what to do about `inherited_members` so I left them alone :)

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -917,7 +917,8 @@ class Module(Doc):
                            for name, obj in inspect.getmembers(self.obj)
                            if (_is_public(name) and
                                is_from_this_module(obj))]
-
+            index = list(self.obj.__dict__).index
+            public_objs.sort(key=lambda i: index(i[0]))
         for name, obj in public_objs:
             if inspect.isroutine(obj):
                 self.doc[name] = Function(name, self, obj)
@@ -1076,29 +1077,30 @@ class Module(Doc):
                 self._context.get(self.name + '.' + _name) or
                 External(name))
 
-    def _filter_doc_objs(self, type: Type[T]) -> List[T]:
-        return sorted(_filter_type(type, self.doc))
+    def _filter_doc_objs(self, type: Type[T], sort=True) -> List[T]:
+        result = _filter_type(type, self.doc)
+        return sorted(result) if sort else result
 
-    def variables(self):
+    def variables(self, sort=True):
         """
-        Returns all documented module-level variables in the module
-        sorted alphabetically as a list of `pdoc.Variable`.
+        Returns all documented module-level variables in the module,
+        optionally sorted alphabetically, as a list of `pdoc.Variable`.
         """
-        return self._filter_doc_objs(Variable)
+        return self._filter_doc_objs(Variable, sort)
 
-    def classes(self):
+    def classes(self, sort=True):
         """
-        Returns all documented module-level classes in the module
-        sorted alphabetically as a list of `pdoc.Class`.
+        Returns all documented module-level classes in the module,
+        optionally sorted alphabetically, as a list of `pdoc.Class`.
         """
-        return self._filter_doc_objs(Class)
+        return self._filter_doc_objs(Class, sort)
 
-    def functions(self):
+    def functions(self, sort=True):
         """
-        Returns all documented module-level functions in the module
-        sorted alphabetically as a list of `pdoc.Function`.
+        Returns all documented module-level functions in the module,
+        optionally sorted alphabetically, as a list of `pdoc.Function`.
         """
-        return self._filter_doc_objs(Function)
+        return self._filter_doc_objs(Function, sort)
 
     def submodules(self):
         """
@@ -1137,6 +1139,8 @@ class Class(Doc):
                        # in Class._fill_inheritance()
                        if (name in self.obj.__dict__ and
                            (_is_public(name) or name == '__init__'))]
+        index = list(self.obj.__dict__).index
+        public_objs.sort(key=lambda i: index(i[0]))
 
         # Convert the public Python objects to documentation objects.
         for name, obj in public_objs:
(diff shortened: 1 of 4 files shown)


pdoc/__init__.py
```diff
@@ -1210,42 +1209,49 @@ def subclasses(self) -> List['Class']:
...
+        Returns an optionally-sorted list of `pdoc.Function` objects that
         represent this class' methods.
         """
         return self._filter_doc_objs(
-            Function, include_inherited, lambda dobj: dobj.method)
+            Function, include_inherited, lambda dobj: dobj.method,
+            sort)
 
-    def functions(self, include_inherited=True) -> List['Function']:
+    def functions(self, include_inherited=True, sort=True,) -> List['Function']:
```
kernc: No trailing comma.

pdoc/__init__.py
```diff
@@ -1132,11 +1132,10 @@ def __init__(self, name, module, obj, *, docstring=None):
         self.doc.update(_var_docstrings(self))
 
         public_objs = [(name, inspect.unwrap(obj))
-                       for name, obj in inspect.getmembers(self.obj)
+                       for name, obj in self.obj.__dict__.items()
```
kernc: The test breaks because of this. The debugger says:

```py
>>> dict(inspect.getmembers(self.obj))['static']
<function B.static at 0x7f61134ac400>

>>> self.obj.__dict__['static']
<staticmethod object at 0x7f6113540d68>)
```

Maybe use `(name, getattr(self.obj, name)) for name in self.obj.__dict__`, but `inspect.getmembers()` looks like it covers a lot of cases. Maybe best to just:
```py
public_objs = ...  # like before
public_objs.sort(key=lambda i, _index=list(self.obj.__dict__).index: _index(i[0]))
```

pdoc/__init__.py
```diff
@@ -914,9 +914,8 @@ def is_from_this_module(obj):
                 return mod is None or mod.__name__ == self.name
 
             public_objs = [(name, inspect.unwrap(obj))
-                           for name, obj in inspect.getmembers(self.obj)
-                           if (_is_public(name) and
-                               is_from_this_module(obj))]
+                           for name, obj in self.obj.__dict__.items()
+                           if _is_public(name) and is_from_this_module(obj)]
```
kernc: Not necessary for modules.

kernc:
Inherited members are last added to `pdoc.Class.doc`, which will make them listed last in the definition order.

nicolaskruchten:
OK, I've made all the requested changes, and the tests are passing! Thanks for the input :)

Please LMK if you want anything else done here.

kernc:
Indeed, for ease of use, please add a template configuration option `sort_identifiers` and pass it to methods.
https://github.com/pdoc3/pdoc/blob/d8602fcf7238de412e8be6e28c31f48f1d858ce3/pdoc/templates/config.mako#L1-L7
https://github.com/pdoc3/pdoc/blob/d8602fcf7238de412e8be6e28c31f48f1d858ce3/pdoc/templates/html.mako#L19-L26

kernc:
Also please add a short unit test.

nicolaskruchten:
I've threaded the config variable to the default template, but after looking at the current set of tests, it's not clear to me where to insert a test for this functionality... There doesn't seem to be a way to test the templates with different values set in `config.mako`?

pdoc/__init__.py
```diff
@@ -917,7 +917,8 @@ def is_from_this_module(obj):
                            for name, obj in inspect.getmembers(self.obj)
                            if (_is_public(name) and
                                is_from_this_module(obj))]
-
+            public_objs.sort(key=lambda i,
+                             _index=list(self.obj.__dict__).index: _index(i[0]))
```
kernc: mypy linter [fails](https://travis-ci.org/pdoc3/pdoc/jobs/493811840#L233) on this lambda.

nicolaskruchten: fixed!

nicolaskruchten:
Added some tests... didn't cover every function I modified but enough to prove the logic works :)

pdoc/__init__.py
```diff
@@ -1137,6 +1139,8 @@ def __init__(self, name, module, obj, *, docstring=None):
                        # in Class._fill_inheritance()
                        if (name in self.obj.__dict__ and
                            (_is_public(name) or name == '__init__'))]
+        public_objs.sort(key=lambda i,
+                         _index=list(self.obj.__dict__).index: _index(i[0]))
```
kernc: Do the clearer `index = ...` trick here as well.

pdoc/__init__.py
```diff
@@ -1210,42 +1214,48 @@ def subclasses(self) -> List['Class']:
...
         represent this class' instance variables. Instance variables
         are those defined in a class's `__init__` as `self.variable = ...`.
         """
         return self._filter_doc_objs(
-            Variable, include_inherited, lambda dobj: dobj.instance_var)
+            Variable, include_inherited, lambda dobj: dobj.instance_var,
+            sort)
 
-    def methods(self, include_inherited=True):
+    def methods(self, include_inherited=True, sort=True,):
```
kernc: Trailing comma. Two more above. Is this muscle memory? :smiley:

pdoc/templates/html.mako
```diff
@@ -305,12 +306,14 @@
         <li>
         <h4><code>${link(c)}</code></h4>
         <%
-            members = c.functions() + c.methods()
+            members = c.functions(sort=sort_identifiers) + c.methods(sort=sort_identifiers)
             if list_class_variables_in_index:
-                members += c.instance_variables() + c.class_variables()
+                members += c.instance_variables(sort=sort_identifiers) + \
+                  c.class_variables(sort=sort_identifiers)
```
kernc: How about wrapping these two calls in parentheses and a hanging indent?

pdoc/test/__init__.py
```diff
@@ -537,6 +537,31 @@ def test_url(self):
         self.assertEqual(f.url(relative_to=c.module), '#example_pkg.D.overridden')
         self.assertEqual(f.url(top_ancestor=1), 'example_pkg/index.html#example_pkg.B.overridden')
 
+    def test_sorting(self):
```
kernc:
```py
@unittest.skipIf(sys.version_info < (3, 6))
```
The tests only pass deterministically on CPython 3.6+.

pdoc/__init__.py
nicolaskruchten: I think it's an autoformatter in my editor :P

nicolaskruchten:
OK should be good to go :)

kernc:
Thank you.

nicolaskruchten:
Awesome, thanks for accepting my input! Just curious: when do you expect to do the next release on PyPI?

kernc:
pdoc3 0.5.3 has been [released](https://pypi.org/project/pdoc3/).

PR #31, opened 2019-02-14
BUG: Make TOC lists multi-column only when wide enough

kernc:
Fixes https://github.com/pdoc3/pdoc/issues/29

### DIFF
--- a/pdoc/templates/css.mako
+++ b/pdoc/templates/css.mako
@@ -134,9 +134,10 @@
     #index h4 + ul {
       margin-bottom:.6em;
     }
-    #index .two-column {
-      column-count: 2;
-    }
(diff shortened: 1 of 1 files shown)


rec:
Onnit! :-)

rec:
All right, I verified that this did work on my project, Python 3.6.6, Mac OS.

Here's what I did.  

First I checked out your repo and ran my code against master with my template, good, then I patched in your pull request, and ran it, with my template, also good.

Then I turned off my template, using your pull request, and still good.  Finally, in your branch I checked out your master, and it went back to being bad, proving that I was actually using the right code, and that your pull request fixes the problem at least for my project!

kernc:
Thanks!

PR #42, opened 2019-03-02
Fix: Relative path for Sub-Modules on Windows

### LINKED ISSUE #41: Sub-Module Link Error Only on Windows
## Issue

The Sub-Modules links on the left sidebar should link to the file; however, on Windows the links are incorrect. On Windows, the path has an incorrect relative (`../`), which isn't present on Mac (I don't have Linux to check, but probably only an issue on Windows).

### On Windows:

linkissue

Links to http://localhost:8080/diagnose.html (`..\diagnose.html`)

badlink

When it should link to http://localhost:8080/bs4/diagnose.html (`./diagnose.html`)

### On Mac

Works as expected:

MacInspect

## Steps to Reproduce

```sh
# # Setup
# git clone ...
# pip install -e .
# pip install beautifulsoup4

# Demo
pdoc bs4 --http :
# or: pdoc bs4 --html
```

### Additional info

- pdoc version: latest as of 01Mar2019: https://github.com/pdoc3/pdoc/commit/1e0608f7d9d71e458ec283eedd75fbb009649196
- OS: (Windows 7, Mac OS Mojave)

> I'm planning on working on a PR/bug fix for this, but wanted to document what I'm seeing first

KyleKing:
Fixes #41 

`startswith('../')` was only catching UNIX-like paths. Replaced with check for both UNIX and Windows paths
Could also be `if url.startswith('../') or url.startswith('..\\'):`

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -833,8 +833,8 @@ class Doc:
 
         # Otherwise, compute relative path from current module to link target
         url = os.path.relpath(self._url(), relative_to.url())
-        # We have one set of '..' too many. Handle Windows and UNIX slashes
-        if re.match(r'^\.\.[\/\\]', url):
+        # We have one set of '..' too many
+        if url.startswith('..' + os.path.sep):
             url = url[3:]
         return url
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -832,9 +832,9 @@ class Doc:
             return '#' + self.refname
 
         # Otherwise, compute relative path from current module to link target
-        url = os.path.relpath(self._url(), relative_to.url())
+        url = os.path.relpath(self._url(), relative_to.url()).replace(path.sep, '/')
         # We have one set of '..' too many
-        if url.startswith('..' + os.path.sep):
+        if url.startswith('../'):
             url = url[3:]
         return url
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -833,8 +833,8 @@ class Doc:
 
         # Otherwise, compute relative path from current module to link target
         url = os.path.relpath(self._url(), relative_to.url())
-        # We have one set of '..' too many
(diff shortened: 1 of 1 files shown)


pdoc/__init__.py
```diff
@@ -833,8 +833,8 @@ def url(self, relative_to: 'Module' = None, *, link_prefix: str = '',
 
         # Otherwise, compute relative path from current module to link target
         url = os.path.relpath(self._url(), relative_to.url())
-        # We have one set of '..' too many
-        if url.startswith('../'):
+        # We have one set of '..' too many. Handle Windows and UNIX slashes
+        if re.match(r'^\.\.[\/\\]', url):
```
kernc: Nice! But maybe just: `url.startswith('..' + os.path.sep)`.

kernc: Which will also make the extra comment redundant.

KyleKing: Works like a charm!

pdoc/__init__.py
```diff
@@ -834,7 +834,7 @@ def url(self, relative_to: 'Module' = None, *, link_prefix: str = '',
         # Otherwise, compute relative path from current module to link target
         url = os.path.relpath(self._url(), relative_to.url())
         # We have one set of '..' too many
-        if url.startswith('../'):
+        if url.startswith('..' + os.path.sep):
```
kernc: Yesterday was a bit late. In fact this is not ok. `url` should not contain Windos path separators. Even if we strip the first one, any additional separators will remain backslashes instead of forward-slashes (URL).

Either `self._url()` needs to be fixed if it returns backslashes, or if backslashes are returned by `os.path.relpath()` call above, they all need to be replaced.

KyleKing: Okay, I think the latest version should address that issue. Tested on Windows 7

kernc:
Thanks a bunch!

Issue #44, opened 2019-03-22
pdoc cli ignoring virtual environment


jlaufersweiler:
### Expected Behavior
Runing pdoc3 cli inside active venv shell session:
(venv) C:\Users\me\dev\fooproject>pdoc --html foo
Documentation should be  generated based on modules active in the virtual environment.  
### Actual Behavior
ModuleNotFoundError and ImportError raised for any modules/packages imported by foo that are not present in the system Python directory.

### Steps to Reproduce

1. Activate virtual environment.
2. run pdoc cli for module that imports resources that are in the virtual environment but are not in the system environment.

### Additional info

- pdoc version:0.5.3
- platform: Windows 10

kernc:
This is weird because pdoc really just imports.

https://github.com/kernc/pdoc/blob/e7868e286c260650daeee478a8b2f1e9ed4c8767/pdoc/__init__.py#L479-L481

Can you test if calling [`importlib.invalidate_caches()`](https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches) above these lines maybe helps?

What is the output of `python -m site`?

jlaufersweiler:
"python -m site?"
Here it is:
```
(venv) C:\Users\jlaufersweiler\dev\project>python -m site
sys.path = [
    'C:\\Users\\jlaufersweiler\\dev\\project',
    'C:\\Users\\jlaufersweiler\\dev\\project\\venv\\Scripts\\python36.zip',
    'C:\\Program Files\\Python36\\DLLs',
    'C:\\Program Files\\Python36\\lib',
    'C:\\Program Files\\Python36',
    'C:\\Users\\jlaufersweiler\\dev\\project\\venv',
    'C:\\Users\\jlaufersweiler\\dev\\project\\venv\\lib\\site-packages',
]
USER_BASE: 'C:\\Users\\jlaufersweiler\\AppData\\Roaming\\Python' (exists)
USER_SITE: 'C:\\Users\\jlaufersweiler\\AppData\\Roaming\\Python\\Python36\\site-packages' (exists)
ENABLE_USER_SITE: False
```
This correctly sees the venv layered atop the system-wide 3.6 install.
In contrast, running pdoc att the same prompt is and searching executing from my user-specific 3.7 instance beneath ~\AppData\Local...:
`(venv) C:\Users\jlaufersweiler\dev\project>pdoc --html batconvert
Traceback (most recent call last):
[...]

PR #45, opened 2019-03-28
FIX: pdoc cli ignoring virtual environment

jlaufersweiler:
fixes: https://github.com/pdoc3/pdoc/issues/44

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -396,25 +396,14 @@ def main(_args=None):
     # Support loading modules specified as python paths relative to cwd
     sys.path.append(os.getcwd())
 
-    # Virtual Environment Handling for system pdoc cli
+    # Virtual environment handling for pdoc script run from system site
     try:
-        # this will raise a KeyError if not in a virtual envitonment.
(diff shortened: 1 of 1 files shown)


pdoc/cli.py
```diff
@@ -299,6 +299,26 @@ def main(_args=None):
...
+                                'lib',
+                                'site-packages',
+                                'pdoc')):
+
+            pass
+        else:
+            # pdoc cli is running with system Python context despite being invoked in a v.e.
+            sys.path.append(path.join(venvPath,
+                                      'lib',
+                                      'site-packages'))
```
kernc: [virtualenv docs say](https://virtualenv.pypa.io/en/latest/userguide/#usage) the path is `ENV/lib/pythonX.X/site-packages/`?

kernc: The path seems to be different on Windos: https://docs.python.org/3/library/venv.html#creating-virtual-environments

kernc:
Applied with slight modification. Thanks for the solution and apologies for the long turnaround time.

PR #47, opened 2019-03-29
Allow for no call to subclasses for a metaclass

mbustosorg:
It looks like a check in pdoc was not migrated to pdoc3?  This is a check to make sure subclasses isn't called since the interface is different.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1210,6 +1210,8 @@ class Class(Doc):
         The objects in the list are of type `pdoc.Class` if available,
         and `pdoc.External` otherwise.
         """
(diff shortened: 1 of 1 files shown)


mbustosorg:
Checking the tests.

kernc:
Why did you change your mind?

-----
Would calling `type.__subclasses__(self.obj)` work better?

mbustosorg:
The tests were failing.  I'm working on fixing them now and will resubmit a
pull request when it's done.

Yes your suggestion is the direction I'm thinking of going.

Thanks for checking and I'll get back to you soon.

mauricio

PR #48, opened 2019-04-01
Support metaclass subclasses

mbustosorg:
Descendants should check if the class is a Metaclass.  Added test also.


pdoc/__init__.py
```diff
@@ -1210,6 +1210,10 @@ def subclasses(self) -> List['Class']:
         The objects in the list are of type `pdoc.Class` if available,
         and `pdoc.External` otherwise.
         """
+        if issubclass(self.obj, type):
+            return [self.module.find_class(c)
+                    for c in self.obj.__subclasses__(self.obj)]
+
```
kernc: I think the whole method can be replaced with:
```py
return [self.module.find_class(c)
        for c in type.__subclasses__(self.obj)]
```
This makes the branch redundant as it works for both cases.

mbustosorg:
d.py is tested as part of the example_pkg test.

Calling with self.obj doesn't work because non metaclass type classes take
no arguments.  The tests fail that way.

Thanks,
mauricio

kernc:
I think that's because `self.obj.__subclasses__` is a bound method whereas `type.__subclasses__` is unbound and can be passed the `self.obj` (meta)class parameter. Seems to work.

mbustosorg:
I tried that iteration and it failed when testing d.py.  Did you see different results?

kernc:
Seems to work here:
```py
>>> class A: pass
>>> class B(type): pass
>>> class C(A): pass
>>> class D(B): pass
>>> type.__subclasses__(A)
[<class '__main__.C'>]
>>> type.__subclasses__(B)
[<class '__main__.D'>]
```

In addition, would you consider creating a new test `ApiTest.test_subclasses()` where you create a few local classes and call `pdoc.Class(cls.__name__, sys, cls).subclasses()` on them?

mbustosorg:
Yes.  I'll have a look at why I'm seeing different results as you and will
create the new test case.

pdoc/test/__init__.py
```diff
@@ -467,6 +467,23 @@ def test_inherited_members(self):
...
+        class B(type):
+            pass
+
+        class C(A):
+            pass
+        
+        class D(B):
+            pass
+        a_subclasses = type.__subclasses__(A)
+        b_subclasses = type.__subclasses__(B)
```
kernc: Let's instead test pdoc's API (`pdoc.Class.subclasses`):
```py
any_module = pdoc.Module(pdoc)
self.assertEqual(pdoc.Class('A', any_module, A).subclasses(), [C])
```

pdoc/test/example_pkg/_test_linking/d.py
```diff
@@ -0,0 +1,10 @@
+class D(type):
+
+    def __init__(cls, name, bases, dct):
+        super().__init__(cls, name, bases, dct)
+        cls._instance = None
+
+    def __call__(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance.__call__(*args, **kwargs)
+        return cls._instance
```
kernc: Afterwards, this file is obsolete.

mbustosorg:
So you would recommend creating a new submodule that includes these classes
instead of in the code like this?

kernc:
No, no, I love how few extra lines are needed to define those classes! Just test _pdoc_ API (`pdoc.Class.subclasses`) instead of Python API (`type` dunder methods). :smile:

kernc:
Sorry, with `any_module` I meant it can be a dummy module just so long the constructor executes.

kernc:
Applied with slight modification. I see I have been oversimplifying my examples. Thanks for figuring it out! :smile:

PR #51, opened 2019-04-26
Change type annotation arrow character

marcwilson:
The new type annotations HTML output use an arrow character that isn't widely supported, or at least it doesn't render for me in Chrome 73 on macOS Mojave.
Can we change to a more commonly supported character, ie: &rarr;

Change from the arrow character used to render the return type annotation from
WIDE-HEADED RIGHTWARDS LIGHT BARB ARROW (http://www.fileformat.info/info/unicode/char/1f862/index.htm)
to the more widely supported
RIGHTWARDS ARROW (https://www.fileformat.info/info/unicode/char/2192/index.htm)

### DIFF
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -90,7 +90,7 @@
             params = ', '.join(f.params(annotate=show_type_annotations, link=link))
             returns = show_type_annotations and f.return_annotation(link=link) or ''
             if returns:
-                returns = ' \u2192\xA0' + returns
+                returns = ' ->\xA0' + returns
         %>
         <span>${f.funcdef()} ${ident(f.name)}</span>(<span>${params})${returns}</span>
     </code></dt>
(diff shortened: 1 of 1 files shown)


kernc:
I've had those concerns. :sweat_smile: Agreed, thanks!

But. The new arrow in the default monospace font looks too tiny. I'd prefer to see it slightly bigger.

Screenshot-2019-4-26 pdoc API documentation

Can you look into it?

marcwilson:
Yes, it does render on the smaller side (I guess that is why they created a bigger one!)
I've tried all the commonly supported right arrows from https://en.wikipedia.org/wiki/Arrow_(symbol) but they all render similarly small in that font for me.

Another option would be to just use the same "->" arrow that is used in the source?
Not as pretty but it would always be the same size as the text in any font

kernc:
Or maybe an arrow styled in CSS?

How does this look to you?
```html
<span style="transform:scale(1.5, 3); display:inline-block; top:-3px; position:relative; margin:0 4px;">→</span>
```

marcwilson:
It looks a little bit weird to me:
image
and feels that all the CSS tweaking values would be specific to the actual font that is being used, so won't be consistent across users.

My preference would just be to go with -> if we want the arrow to be more pronounced as it will just work in all cases, albeit not as pretty:
image

kernc:
:+1: then to change to `->`. It'll look nice once the mainstream monospace fonts support common programming ligatures.

PR #57, opened 2019-05-01
move main docstring into a file

cs01:
This makes the code easier to read since it's not underneath a large docstring. It also makes the documentation easier to edit since text editors can add syntax highlighting to the raw markdown content, and do realtime markdown previews.


pdoc/__init__.py
```diff
@@ -1,345 +1,5 @@
 """
-Python package `pdoc` provides types, functions, and a command-line
-interface for accessing public documentation of Python modules, and
-for presenting it in a user-friendly, industry-standard open format.
-It is best suited for small- to medium-sized projects with tidy,
-hierarchical APIs.
```
kernc: Literate code; leave the first paragraph in.

pdoc/documentation.md
```diff
@@ -0,0 +1,335 @@
...
+  `:start-line:`, `:end-line:`, `:start-after:` and `:end-before:`.
+* `.. versionadded::`
+* `.. versionchanged::`
+* `.. deprecated::`
+* `.. todo::`
+
+[admonitions]: http://docutils.sourceforge.net/docs/ref/rst/directives.html#admonitions
+[image]: http://docutils.sourceforge.net/docs/ref/rst/directives.html#images
+[include]: \
+http://docutils.sourceforge.net/docs/ref/rst/directives.html#including-an-external-document-fragment
```
kernc: While inside python string, the backslash escaped trailing \n. Now in this markdown document, I guess it's a verbatim backslash. These lines now won't be checked by flake8, so I guess it's ok to put on a single line.

cs01:
I want to run the markdown autoformatter on this file to make it more consistent, but I can't tell what format its contents are. Is it markdown? Markdown with some pdoc syntax mixed in? Rst?

pdoc/documentation.md
```diff
@@ -0,0 +1,334 @@
...
+By convention, variables defined in a class' `__init__` method
+and attached to `self` are considered and documented as
+instance variables.
+
+Class and instance variables can also [inherit docstrings].
+
+[inherit docstrings]: #docstrings-inheritance
+
+
+Overriding docstrings with `__pdoc__` asdf
```
kernc: asdf

kernc:
I'm sorry for not having been more verbose up front. Let's just do the moving in this PR (which will become a commit) and the TBD reflowing in a subsequent PR. I'm sure you can see the advantage of this (large, strictly moving remaining a stand-alone commit).

kernc:
> make it more consistent

What do you mean? :confused: 

It's actually simple Markdown Extra. Really just what is supported by Python Markdown with enabled [plugins](https://python-markdown.github.io/extensions/#officially-supported-extensions).
https://github.com/pdoc3/pdoc/blob/eda27f3d003fbbb01adf1008bd21ef69f3868a02/pdoc/html_helpers.py#L67-L77

kernc:
> It's actually simple Markdown Extra.

And some reST directives, which are in the [pipeline](https://github.com/pdoc3/pdoc/blob/eda27f3d003fbbb01adf1008bd21ef69f3868a02/pdoc/html_helpers.py#L189-L238) transformed into markdown [admonitions](https://python-markdown.github.io/extensions/admonition/).

kernc:
Btw, thanks for this! Indeed, both files are much more manageable now! :smile: :rocket:

cs01:
Ah sorry about that "revert typo" commit. What happened there was I am using VS Code and they have an auto-formatter for markdown files which I forgot I ran. 

I was asking earlier about the language of the docstring, because when I ran the VS Code autoformatter it was stripping away what looked like useful information. I see now that is because pdoc is using various syntax that the autoformatter doesn't recognize.

> >make it more consistent

> What do you mean? 

Consistent heading syntax, for example. i.e. `##` next to title vs `------` underneath the title. Also various capitalization, etc. See the changes in [this diff](https://github.com/pdoc3/pdoc/pull/57/commits/5d2ced5264ee9e58b9d5b3c8749c7503a6cb9fc8).

```
 Where does `pdoc` get documentation from?	
-----------------------------------------
```
is transformed to 
```
## Where does `pdoc` get documentation from?
```

kernc:
This PR introduced a bug with sidebar TOC no longer rendering when the docstring was included from another file. @cs01 Looking at your website actually made me aware of it. :smile: It's been fixed in 54fc6d04380ac3fbc36f951cb4842b2903ed53be, and the whole lot released moments ago as v0.6.0. Thanks again!

cs01:
Nice, thanks for the release!

PR #59, opened 2019-05-01
use consistent header syntax

cs01:


kernc:
I'm sorry, I can't accept this. :sweat_smile: Besides bikeshedding, underlined headers are "consistent" syntax in Markdown (`=====` for H1, `------` for H2), they take 2 lines just the same, but are likely much easier to spot in plain text source.

Issue #60, opened 2019-05-02
live reload when serving locally


cs01:
https://twitter.com/burntsushi5/status/1124671095725002752

kernc:
While the web page needs to be refreshed manually, the server doesn't (shouldn't) require to be shut down and restarted as module reloading is in place.
https://github.com/pdoc3/pdoc/blob/f5b4a5484a2add2ab1b276267248fe54e1dbf1a0/pdoc/__init__.py#L169-L175

If there's a bug somewhere, the `import_module()` mechanism is being improved as we speak.

cs01:
Ah I was editing docs for pdoc itself, hence the server wasn't reloading. Still the client-side reload is a nice feature. What do you think?

kernc:
If something can be done in about three lines of WebSocket JS API and maybe twice that on the Python end without heavy extra dependencies, absolutely love to see it.

PR #65, opened 2019-05-07
live reload when serving locally #60

cs01:
This adds the live reload functionality.

> If something can be done in about three lines of WebSocket JS API and maybe twice that on the Python end without heavy extra dependencies, absolutely love to see it.

It takes zero lines of JS (!) and ended up with approximately the same lines of Python. It uses a single dependency, `livereload`.

Watches files for changes, then automatically regenerates html and reloads any webpages that are viewing the generated html. This works on pdoc as well now.

TODO
- [x] Get proof of concept working
- [ ] Honor other pdoc arguments (I didn't look into them in detail, because I didn't want to spend too much time on this before I got initial feedback from you)
- [ ] Call a pdoc function rather than shelling out to a subprocess
- [ ] Generate index.html to avoid needing to go to `host:port/module`
- [ ] Update unit tests

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -9,7 +9,9 @@ import os.path as path
 import re
 import sys
 import warnings
-from http.server import BaseHTTPRequestHandler, HTTPServer
+from livereload import Server, shell
+from http.server import BaseHTTPRequestHandler
+from tempfile import TemporaryDirectory
 from typing import Sequence
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


kernc:
Have a look at https://github.com/pdoc3/pdoc/pull/66 as an alternative. It's double the anticipated lines of JS, but it reuses existing Python server architecture (behavior), fixes some legacy bugs, and introduces no new dependencies.

PR #68, opened 2019-05-13
Fix generating Markdown when using outputdir

Aurora-Beta:
The generation of textfiles / markdownfiles was not working when the parameter -o / --outputdir was omitted, instead, HTML-Files were generated.


kernc:
Thanks! :sweat_smile:

PR #72, opened 2019-06-21
dum bug with doctests were getting clipped bc of regex

jessekrubin:
doctests were being clipped in the html => markdown conversion. one too many new lines in the rejects (regex)

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -291,7 +291,7 @@ class _ToMarkdown:
                      re.compile(r'(?:^(?P<fence>```|~~~).*\n)?'
                                 r'(?:^>>>.*'
                                 r'(?:\n(?:(?:>>>|\.\.\.).*))*'
-                                r'(?:\n.*)?\n\n?)+'
+                                r'(?:\n.*)?\n?)+'
                                 r'(?P=fence)?', re.MULTILINE).sub,
                      lambda m: (m.group(0) if m.group('fence') else
                                 ('\n    ' + '\n    '.join(m.group(0).split('\n')) + '\n\n')))):


kernc:
Agreed, regexes aren't the most robust method of transforming arbitrary markup. But can you maybe [show an example issue](https://github.com/pdoc3/pdoc/issues) this PR fixes?

jessekrubin:
> Agreed, regexes aren't the most robust method of transforming arbitrary markup. But can you maybe [show an example issue](https://github.com/pdoc3/pdoc/issues) this PR fixes?

# __PROBLEM__

problem


# __UNPROBLEM__
unproblem

kernc:
Trying to come up with a test case, but can't seem to reproduce the issue. :confused: 

This correctly indents the _whole_ code block:
```py
text = '''
this function will convert bytes to MB.... GB... etc

Examples
--------
>>> nbytes(100)
'100.0 bytes'
>>> nbytes(1000)
'1000.0 bytes'
>>> nbytes(10000000000000)
'9.1 TB'
>>> nbytes(100000000000000)
'90.9 TB'
'''
from pdoc.html_helpers import _ToMarkdown
print(_ToMarkdown.doctests(text))
```
```text
this function will convert bytes to MB.... GB... etc

Examples
--------

    >>> nbytes(100)
    '100.0 bytes'
    >>> nbytes(1000)
    '1000.0 bytes'
    >>> nbytes(10000000000000)
    '9.1 TB'
    >>> nbytes(100000000000000)
    '90.9 TB'
```

jessekrubin:
Here's my dummy test case

```    
def nbytes(num) -> str:
    """
    this function will convert bytes to MB.... GB... etc

    Examples:

        >>> nbytes(100)
        '100.0 bytes'
        >>> nbytes(1000)
        '1000.0 bytes'
        >>> nbytes(10000)
        '9.8 KB'
        >>> nbytes(100000)
        '97.7 KB'
        >>> nbytes(1000000)
        '976.6 KB'
        >>> nbytes(10000000)
        '9.5 MB'
        >>> nbytes(100000000)
        '95.4 MB'
        >>> nbytes(1000000000)
        '953.7 MB'
        >>> nbytes(10000000000)
        '9.3 GB'
        >>> nbytes(100000000000)
        '93.1 GB'
        >>> nbytes(1000000000000)
        '931.3 GB'
        >>> nbytes(10000000000000)
        '9.1 TB'
        >>> nbytes(100000000000000)
        '90.9 TB'

    """
    for x in ["bytes", "KB", "MB", "GB", "TB"]:
        if num < 1024.0:
            return "%3.1f %s" % (num, x)
        num /= 1024.0

```

kernc:
Thanks for the report! My thinking is that ensuring line termination seems better than changing the regex, particularly as I don't understand it. :clown_face:

Issue #70, opened 2019-06-08
Mathematics in docstrings ?


lmoresi:
We (@underworldcode) recently adopted pdoc3 for some of our pure python projects and we were wondering about the status of mathematical equations in the generated documentation. 

We often include snippets of latex in our docstrings for cases where we are implementing something overtly mathematical that needs explaining and where an explanation that tries to skirt the equations is horribly convoluted. 

An example here: http://www.underworldcode.org/stripy/cartesian.html#stripy.cartesian.Triangulation.interpolate_cubic. The docs in this case are from pdoc3 with a tweak to add mathjax headers and include the markdown mathematics extensions. 

We'd be interested to know if this is a benign addition to pdoc3 that could ultimately be useful / optionally supported or if it is a niche case of little interest in real life.

kernc:
This certainly should be supported. :+1: The preferred mechanism of MathJax inclusion and configuration is by overriding [_head.mako_ template](https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates). How else are you doing it?

lmoresi:
Thanks for responding.

Over-riding the template is the first part, but there are two issues that caused me to raise this with the devs as a question.

The first is that it is helpful (but not essential) to add the "mdx_math" markdown extension that wraps all the math tags nicely but then adds a dependency not everyone needs - so I wasn't sure how you would approach that without hacking into the code.

The second is that latex uses \ in nearly every command and this triggers a lot of deprecated escape character warnings from pdoc3.  This can be avoided by escaping all the backslashes and writing equations as $$\\sin^2 \\theta + \\cos^2 \\theta = 1$$ (that's what we did in our example docs linked above) but that looks weird to a latex person. The other way is to treat/mark all docstrings as raw. If we mark them as raw then no escapes are necessary, but the code looks weird to a python person. I don't know what side effects that has in more general pdoc3 usage or with docstrings as they appear in notebooks. 

In other words, our modifications work for our very specific project and we don't have the experience to see if there are problems ahead. We thought pdoc3 was amazing - up and running in just a few minutes and with very clear coding style behind the scenes - so we'd like to adopt it as our default and decide on a "house-style" for our equation markup.

Thanks for your time.

L

kernc:
Sorry for the late reply.

[...]

PR #74, opened 2019-06-23
ENH: Support LaTeX math syntax with MathJax

kernc:
Fixes https://github.com/pdoc3/pdoc/issues/70.

Includes new `latex_math=False` config option.


lmoresi:
That seems to work very well. Thanks for that. I have tested it in the limited case of our documentation and it does exactly what we had hacked some cheap code to achieve. I'm happy to defer to your expertise regarding the markdown math extensions.

kernc:
Your mockery was well-placed. :sweat_smile: I opted for Python-Markdown Extension API anyway. I'm sure @waylan goes out of his way to make the API more stable and well-supported than any hackery I could come up with. :+1:

PR #77, opened 2019-07-04
FIX: Handle ro-value-descrptiors (#76)

### LINKED ISSUE #76: Support for read-only value descriptors
### Expected Behavior

Detect if read-only descriptor is data or method descriptor.

Generate documentation.

### Actual Behavior

```
 File "/HOME/.local/lib/python3.7/site-packages/pdoc/__init__.py", line 1020, in __init__
    assert callable(obj)
```

in `Class.__init__` (L822) `inspect.isroutine(obj)` is checked and in `Function.__init__` (L1020) `callable(obj)` is asserted.

Problem is when `obj` is read-only value descriptor (has `__get__`, no `__set__` and `__get__()` returns non-callable value).

From `inspect.py` (Python 3.7):
```python
def isroutine(object):
    """Return true if the object is any kind of function or method."""
    return (isbuiltin(object)
            or isfunction(object)
            or ismethod(object)
            or ismethoddescriptor(object))
```

Follow documentation `inspect.ismethoddescriptor(object)`: 
> An object passing this test has a `__get__()` method but not a `__set__()` method...

From  `inspect.py` (Python 3.7):
```python
def ismethoddescriptor(object):
    if isclass(object) or ismethod(object) or isfunction(object):
        # mutual exclusion
        return False
    tp = type(object)
    return hasattr(tp, "__get__") and not hasattr(tp, "__set__")
```

Then assert failed on read-only value descriptors (like `peewee.BackrefAccessor`) and documentation is not generated.

### Steps to Reproduce

1. Use read-only value descriptor, for example (`foo.py`):
```python
class Power:
    def __get__(self, instance, instance_type=None):
        if instance is not None:
            return instance.value ** 2
        return self

class Value:
    power = Power()
    def __init__(self, value):
        self.value = value
```

2. Run `pdoc --html foo.py`

### Additional info

- pdoc version: 0.6.3

- minimal workaround I've found, change `Class.__init__()` (L822):
```diff
--- pdoc-0.6.3/__init__.py  2019-07-04 07:34:53.857974401 +0200
+++ pdoc/__init__.py  2019-07-04 07:34:33.366110353 +0200
@@ -819,7 +819,7 @@
         for name, obj in public_objs:
             if name in self.doc and self.doc[name].docstring:
                 continue
-            if inspect.isroutine(obj):
+            if inspect.isroutine(obj) and callable(obj):
                 self.doc[name] = Function(
                     name, self.module, obj, cls=self,
                     method=not self._method_type(self.obj, name))
```

rysson:
### FIX: Handle read-only value descriptors.

Fixes https://github.com/pdoc3/pdoc/issues/76
Fixes https://github.com/pdoc3/pdoc/issues/88

Fixes assert when class uses read-only value descriptors.
```
 File "/HOME/.local/lib/python3.7/site-packages/pdoc/__init__.py", line 1020, in __init__
    assert callable(obj)
```

### Steps to Reproduce

1. Use read-only value descriptor, for example (`foo.py`):
```python
class Power:
    def __get__(self, instance, instance_type=None):
        if instance is not None:
            return instance.value ** 2
        return self

class Value:
    power = Power()
    def __init__(self, value):
        self.value = value
```

2. Run `pdoc --html foo.py`


kernc:
Should there be another `elif inspect.ismethoddescriptor(obj)` block that inits
`Variable(..., obj=getattr(obj, '__get__'))` for later calls to `Doc.source` to work?

https://github.com/pdoc3/pdoc/blob/366dc1a2f7992e5fcee68c3dbd99e4f73a2c95ce/pdoc/__init__.py#L826-L832

kernc:
Can you also add a test? I think something simple like `pdoc.test.ApiTest.test_instance_var` should be enough.

https://github.com/pdoc3/pdoc/blob/366dc1a2f7992e5fcee68c3dbd99e4f73a2c95ce/pdoc/test/__init__.py#L411-L415

rysson:
Good point. I'm not sure, I've found pdoc yesterday :-)

There are two code paths:
1. documented descriptor variable (parsed by `_var_docstrings()`)
2. is undocumented descriptor variable (lines changed in this pull-request)

Read-only descriptor was in class variable section. I've move it to instance in both cases.
There is one extra condition now, I doesn't look pretty:
https://github.com/pdoc3/pdoc/blob/9aada78d30be140dc18b0d298756f09e720334c1/pdoc/__init__.py#L819-L824

I've add simple two tests new `elif` (with `__get__`) as you suggested.

pdoc/__init__.py
```diff
@@ -818,11 +818,19 @@ def __init__(self, name, module, obj, *, docstring=None):
         # Convert the public Python objects to documentation objects.
         for name, obj in public_objs:
             if name in self.doc and self.doc[name].docstring:
+                if inspect.isroutine(obj) and not callable(obj):
+                    assert isinstance(self.doc[name], Variable)
+                    self.doc[name].instance_var = True
```
kernc: > I've move it to instance in both cases. There is one extra condition now, I doesn't look pretty

Agree this is not pretty. And not all descriptors are [necessarily](https://docs.python.org/3/howto/descriptor.html#static-methods-and-class-methods) instance variables anyway?
.

pdoc/test/example_pkg/__init__.py
```diff
@@ -20,6 +20,15 @@ def foo(env=os.environ):
     pass
 
 
+class ReadOnlyValueDescrpitor:
```
kernc: Descriptor.

pdoc/__init__.py
rysson: I agree, not all descriptors. It's why I added `not callable(obj)`.

I've found:
> If an object defines both `__get__()` and `__set__()`, it is considered a data descriptor. Descriptors that only define `__get__()` are called non-data descriptors (they are **typically** used for methods but **other uses are possible**).

This is a problem. I try to handle non-typical (non-)data descriptor. It's not my idea, I found that in peewee. I think it's better than assertion fail. 
Next, it is better if that type of descriptor will be detected independently it's documented on not.
* Documented variables are found in `_var_docstrings()`, I don't how to detect there almost-non-data descriptors, and set `instance_var` to true there. I change it in the *the ugly code* instead.
* Undocumented are handled now in (after `if inspect.isroutine(obj) and callable(obj)` fails):
https://github.com/pdoc3/pdoc/blob/9aada78d30be140dc18b0d298756f09e720334c1/pdoc/__init__.py#L829


How should it be handle in your opinion?

P.S. Thanks for point me typos.

pdoc/test/__init__.py
```diff
@@ -414,6 +414,23 @@ def test_instance_var(self):
...
+        mod = pdoc.Module(pdoc.import_module(EXAMPLE_MODULE))
+        var = mod.doc['B'].doc['squere']
+        self.assertIsInstance(var, pdoc.Variable)
+        self.assertTrue(var.instance_var)
+        self.assertEqual(var.docstring, """Squere of variable `var`""")
+
+    def test_readonly_var_descriptors_no_docstring(self):
+        pdoc.reset()
+        mod = pdoc.Module(pdoc.import_module(EXAMPLE_MODULE))
+        var = mod.doc['B'].doc['squere_no_doc']
```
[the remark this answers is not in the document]
rysson: OK. I'm merging two cases in one method.
I'm not sure about `var.source`. It's true only for non-documented variable and I still don't why.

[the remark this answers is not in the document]
kernc: `var.source` doesn't evaluate true for the PEP224-documented member because that one is discovered inside `_var_docstrings()` which parses the AST and doesn't produce "live" objects, hence `Doc.obj` property is None and `getsource` fails.
https://github.com/pdoc3/pdoc/blob/a4a9a22056e8c2182ff5ea8d65b4718ee79439bc/pdoc/__init__.py#L428-L431

pdoc/__init__.py
```diff
@@ -818,11 +818,19 @@ def __init__(self, name, module, obj, *, docstring=None):
...
                 continue
-            if inspect.isroutine(obj):
+            if inspect.isroutine(obj) and callable(obj):
```
kernc: How about create new local functions `is_function()` and `is_data_descriptor()` above the for loop and use those in this if conditions. Makes it more verbose.

pdoc/__init__.py
kernc: I wasn't convinced PEP224-documented non-callable descriptors should always be marked as instance variables (referring to `instance_var = True` override), but I guess that's more what they end up being used for.

What breaks if, instead of the override, you move the `self.doc.update(_var_docstrings(self))` below the loop, i.e. add the PEP224-documented variables afterwards?

stevegore:
Thanks for this! Fixed my issue with a read-only property.

rysson:
@kernc, you are great!
Vacation and ... totally overtime work after. I'm sorry.

kernc:
No problem. Thanks for your work on this. :+1:

PR #83, opened 2019-07-23
Allow compatibility with pre-commit

lamourj:
Adding the config file would allow to simply call pdoc on a [pre-commit hook](https://github.com/pre-commit/pre-commit).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+-   id: pdoc
+    name: pdoc
+    description: 'pdoc: Auto-generate API documentation for Python projects'
+    entry: pdoc
+    language: python
+    language_version: python3
+    require_serial: true
+    types: [python]


lamourj:
CI should not fail; it fails for the same reason as the previous merged commits (such as c48c722).
See issue [mypy#4288](https://github.com/python/mypy/issues/4288).

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,8 @@
+-   id: pdoc
+    name: pdoc
```
kernc: At least some of these `pdoc` tags (not sure which, probably all) need to be changed to `pdoc3` if pre-commit fetches Python packages from PyPI.

kernc: One actually [specifies their own hook source repo](https://pre-commit.com/#2-add-a-pre-commit-configuration). :sweat_smile:

kernc:
Thanks! And sorry for the long turnaround time.

PR #93, opened 2019-08-16
Bug Fix: Producing markdown output currently fails

timothycrosley:
Fix incorrect variable reference when producing mark down output and having type annotation


kernc:
Thanks!

PR #94, opened 2019-08-17
DOC: mention show_source_code=False in Examples

zachvalenta:
I never come across this flavor of Markdown before, so let me know if I'm doing anything that's not idiomatic (or just plain wrong).


kernc:
I meant more like this. :grin:  Thanks!

PR #103, opened 2019-08-29
Refactor args.config processing

dhimmel:


pdoc/cli.py
```diff
@@ -370,12 +371,17 @@ def main(_args=None):
...
+    template_config = {}
+    for opt in args.config:
+        try:
+            config_key, config_value = opt.split('=', 1)
+            config_value = ast.literal_eval(config_value)
+            template_config[config_key] = config_value
+        except Exception as e:
+            raise RuntimeError(
+                'Error evaluating config value {!r}\n'
+                'Make sure string values are quoted?'.format(opt, e)
```
kernc: Template is missing a placeholder for `e`.

If `opt` is repr-ed, don't we end up with too many quotes?

dhimmel: > If opt is repr-ed, don't we end up with too many quotes?

Possibly, it could be useful for the `unquoted configuration value` case, but probably better without overall.

```python
>>> print("{!r}".format("'configuration_value'"))
"'configuration_value'"
>>> print("{!r}".format("'configuration value'"))
"'configuration value'"
>>> print("{!r}".format('"configuration value"'))
'"configuration value"'
>>> print("{!r}".format("unquoted configuration value"))
'unquoted configuration value'
```

dhimmel:
As of https://github.com/pdoc3/pdoc/pull/103/commits/78d31523b2b2348e74af8c33e28cb0075df757f4, here is the output for a bad --config:

```python-traceback
pdoc3 --html \
  --template-dir doc/pdoc_template \
  --force --output-dir doc \
  --config "variable=this should be bad" \
  pdoc

  File "/home/dhimmel/Documents/repos/pdoc/pdoc/cli.py", line 378, in main
    config_value = ast.literal_eval(config_value)
  File "/home/dhimmel/anaconda3/lib/python3.6/ast.py", line 48, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/home/dhimmel/anaconda3/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    this should be bad
              ^
SyntaxError: invalid syntax

  File "/home/dhimmel/Documents/repos/pdoc/bin/pdoc3", line 11, in <module>
    load_entry_point('pdoc3', 'console_scripts', 'pdoc3')()
  File "/home/dhimmel/Documents/repos/pdoc/pdoc/cli.py", line 384, in main
    .format(error.__class__.__name__, config_str, error)
RuntimeError: SyntaxError evaluating --config variable=this should be bad
invalid syntax (<unknown>, line 1)
Make sure string values are quoted?
```

A bit repetitive and verbose, but it gets the problem across. Another option is to raise a warning rather than an error and proceed with execution.

pdoc/cli.py
dhimmel: > Template is missing a placeholder for `e`.

Addressed in https://github.com/pdoc3/pdoc/pull/103/commits/78d31523b2b2348e74af8c33e28cb0075df757f4

kernc:
Too severe for a warning. Made it just a tad less repetitive, as you say. Thanks!

Issue #101, opened 2019-08-28
Index of documented modules


ryangalamb:
### Expected Behavior
A way to include the `--http` mode's "Python module list" `index.html` as the root `index.html` of the `--html` output. 

### Actual Behavior
By default, there is no `index.html` at the top level of the HTML output. There doesn't seem to be a way to enable the "Python module list" in the static html output.

### Steps to Reproduce

1. `pdoc --html [some modules]`
2. nothing at `./html/index.html`

### Additional info

I'm happy to open a PR for this if it seems like a reasonable request. I'm imagining it as a command line flag to run in addition to `--html`. When present, pdoc would generate an index.html with the "Python module list" the same as the one generated in `WebDoc.do_GET()` when path is `"/"`

- pdoc version: pdoc 0.6.3

kernc:
Ref: https://github.com/pdoc3/pdoc/issues/95.

I think it's safe to just generate the index.html always, why not? Didn't think of that in https://github.com/pdoc3/pdoc/issues/55.

Besides the index file, why not list all top-level, disjunct packages/modules specified on a single command line also in the sidebar below the members list? :thinking:

rjmill:
@kernc 
>I think it's safe to just generate the index.html always, why not?

I'd be hesitant to turn this feature on by default. For my own use case, my docs automation would break if we made this the default. I'm sure others would experience similar problems.

Also, if we make this feature opt-in, we can do nice things like enabling the module breadcrumbs that show up when users are running in `--http` mode:
> ![image](https://user-images.githubusercontent.com/12422195/63981297-7e9d9700-ca8c-11e9-9efa-a48cd89b843b.png)

That feature wouldn't make sense for users who aren't going to be using the root `index.html` though.

I'll open a WIP PR to show you what I've got so far and what I'm thinking.

kernc:
I'm hesitant adding command line flags for little used features.

What do you need the index for and how often? Couldn't a simple shell script, iterating over the same packages specified on the command line, do the job?
```bash
packages="package1 package2 package3"
for pkg in $packages; do echo "<a href='$pkg'>$pkg</a><br>" >> index.html; done
```

What if the index were only created if the user specifies more than one package?

kernc:
I think, rather in factly, that a proper index is indeed in order, *spanning hierarchically all the way to leaf modules*.

Created always.

[...]

PR #104, opened 2019-08-29
ENH: include root `index.html` in `--html` output

rjmill:
Implements #101 .

It adds the root `index.html` from the `--http` mode to the static `--html` output.

To avoid breaking people's automations, I've put this feature behind a flag (`--html-index`) and left the default behavior alone.

I also added the breadcrumbs to the module pages when `--html-index` is active.

You can use as much or as little of this as you want. I did this in several small commits so you can see all the options side-by-side.  

**What's left**
- [x] put the feature behind a flag
- [x] check for file existence like the rest of the files do
- [x] write tests for the new flag
- [x] include breadcrumbs

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -487,6 +487,25 @@ or similar, at your own discretion.""",
             # Two blank lines between two modules' texts
             sys.stdout.write(os.linesep * (1 + 2 * int(module != modules[-1])))
 
+    # Add the root index.html at the top level
+    if args.html:
+        # The template expects `modules` to be Tuples of (name, docstring)
+        module_tuples = sorted((module.name, module.docstring)
+                               for module in modules)
+        index_text = pdoc._render_template('/html.mako',
+                                           modules=module_tuples,
+                                           **template_config)
+        index_file = path.join(args.output_dir, 'index.html')
+        try:
+            with open(index_file, 'w+', encoding='utf-8') as w:
+                w.write(index_text)
+        except Exception:
+            try:
+                os.unlink(index_file)
+            except Exception:
+                pass
+            raise
+
 
 if __name__ == "__main__":
     main(parser.parse_args())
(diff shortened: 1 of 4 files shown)


pdoc/cli.py
```diff
@@ -487,6 +487,25 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
             # Two blank lines between two modules' texts
             sys.stdout.write(os.linesep * (1 + 2 * int(module != modules[-1])))
 
+    # Add the root index.html at the top level
+    if args.html:
```
rjmill: This is adapted from `WebDoc.do_GET` on `"/"`. That one doesn't use the `pydoc.Module` class for the root index. It uses `args.modules` and imports the modules to get the name and docstring.

When I tried doing it in the same way here, there were some inconsistencies with the rest of the `--html` behavior that was causing some weird breakages and test errors. Otherwise, I would have pulled the logic out into a helper function.

pdoc/cli.py
```diff
@@ -487,6 +487,25 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
...
+        # The template expects `modules` to be Tuples of (name, docstring)
+        module_tuples = sorted((module.name, module.docstring)
+                               for module in modules)
+        index_text = pdoc._render_template('/html.mako',
+                                           modules=module_tuples,
+                                           **template_config)
+        index_file = path.join(args.output_dir, 'index.html')
+        try:
```
rjmill: And this logic is adapted from `write_files`. I couldn't use it directly, since it expects a `pdoc.Module` as input, so I just copy-pasted. I can go back and refactor that to be more DRY if you prefer.

pdoc/cli.py
```diff
@@ -487,6 +492,25 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
             # Two blank lines between two modules' texts
             sys.stdout.write(os.linesep * (1 + 2 * int(module != modules[-1])))
 
+    if args.html and args.html_index:
```
rjmill: I'll name the arg whatever you want, but this is the general idea.

rjmill:
To my point about breaking automation: The CI build was failing for "Deploy" with my initial PR, but it's passing now that I've put the feature behind a flag.

kernc:
Let's discuss this back in https://github.com/pdoc3/pdoc/issues/101.

PR #119, opened 2019-10-30
Update command line example

NicolasCARPi:
--output-dir doesn't exist anymore, replace it with --html-dir

### DIFF
--- a/pdoc/documentation.md
+++ b/pdoc/documentation.md
@@ -237,7 +237,7 @@ For example, to produce HTML documentation of your whole package
 in subdirectory 'build' of the current directory, using the default
 HTML template, run:
 
-    $ pdoc --html --output-dir build my_package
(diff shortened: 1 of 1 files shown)


kernc:
I think it's [the other way around](https://github.com/pdoc3/pdoc/blob/1b644f6e86a5da4922679a8db540b83f8362ef08/CHANGELOG#L60). :smile:

NicolasCARPi:
Well, the `--help` doesn't show any `--output-dir` option. Anyway, the doc is wrong for a new user that wants to try it, it'll not work as it is.

kernc:
Have you confirmed you're using the latest version?

NicolasCARPi:
In fact, no, I installed it with pip and it's 0.3.2 there! https://pypi.org/project/pdoc/

That explains! :)

PR #128, opened 2019-11-20
Sort subclasses for cross-platform consistency

pkazmier:
The sort order for subclasses is inconsistent between platforms. As a
result, projects that include generated docs in their repos see diffs to
the generated files based on the platform of the contributor. This PR adds
an explicit sort so we have a deterministic order of subclasses.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -476,7 +476,7 @@ class Doc:
         return top
 
     def __lt__(self, other):
-        return self.name < other.name
+        return self.refname < other.refname
 
 
 class Module(Doc):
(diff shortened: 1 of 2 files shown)


pdoc/__init__.py
```diff
@@ -882,8 +882,9 @@ def subclasses(self) -> List['Class']:
         The objects in the list are of type `pdoc.Class` if available,
         and `pdoc.External` otherwise.
         """
-        return [self.module.find_class(c)
-                for c in type.__subclasses__(self.obj)]
+        return sorted((self.module.find_class(c)
+                       for c in type.__subclasses__(self.obj)),
+                      key=lambda x: x.refname)
```
kernc: `pdoc.Doc` already defines `__lt__()`. Can we modify that to use `pdoc.Doc.refname` canonically when sorting? No tests fail with the change. :grin:

pkazmier: Sure, PR has been updated.

kernc:
Thanks!

Issue #124, opened 2019-11-15
support unusual constructors by using inspect.signature() on class directly


programmerjake:
I'm currently writing a Python extension in Rust using [PyO3](https://pyo3.rs/). PyO3 uses `__new__` instead of `__init__` due to peculiarities needed by Rust's memory safety model.

It would be nice to support documenting the constructor signature by using `inspect.signature()` on the class directly rather than on the `__init__` or `__new__` method, since that also picks up `cls.__text_signature__` derived from the class's documentation.

kernc:
This is the class signature entry point:
https://github.com/pdoc3/pdoc/blob/3a5e1d1241562691cb64584add3c8ed647ea2505/pdoc/__init__.py#L888-L903
If the line is changed to:
```py
params = Function._params(self.obj, ...)
```
and the redundant line after is removed, it all still seems to work and looks cleaner.

Can you propose a simple test case?

programmerjake:
Here's a relatively simple test case:
https://github.com/programmerjake/rust-python-test/tree/doc-test

Build instructions (tested on Ubuntu 18.04 x86_64, in CPython 3.6 venv):
Install Rust toolchain:
```sh
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
Restart terminal window to put Rust in `$PATH` (added to `.bashrc`).

Install build tool:
```sh
pip install maturin
```
Checkout code:
```sh
git clone -b doc-test https://github.com/programmerjake/rust-python-test.git
cd rust-python-test
```
Switch Rust to use known-working nightly version in `rust-python-test` directory:
```sh
rustup override set nightly-2019-07-19
```
Build and install (like setup.py develop):
```sh
maturin develop
```
Generate docs using pdoc3
```sh
pdoc rust_python_test
```

programmerjake:
If you're not ok with using Rust, I can try to write a C version, but it may take some time as I'm not familiar with writing CPython extensions in C.

programmerjake:
I'm not aware of a way to create the correct classes using pure Python.

programmerjake:
Built wheels and source for convenience: [rust_python_test_wheels.zip](https://github.com/pdoc3/pdoc/files/3863215/rust_python_test_wheels.zip)

kernc:
> I'm not aware of a way to create the correct classes using pure Python.

I'm so sorry for the effort you needlessly put into it above. Pure Python test case is indeed the only one I'd consider.

[...]

PR #135, opened 2019-12-10
add test case for #124

programmerjake:
Add test case for #124

I also auto-formatted the code; if you like, I can split that out into a separate commit or leave the code unformatted.


kernc:
Nice hacking. Much thanks!

Issue #99, opened 2019-08-24
Ignore entire package / directory?


amchang:
Hello there,

I actually download my dependencies locally into a subfolder within my package similar to something like this.

```
package/
     __init__.py
     code.py
     downloaded_modules/
           requests/
           ...
```
Is there any way to effectively ignore `downloaded_modules`?

Thanks

kernc:
Exclude items using [`__pdoc__` dict](https://pdoc3.github.io/pdoc/doc/pdoc/#overriding-docstrings-with-__pdoc__).

Duplicate of https://github.com/pdoc3/pdoc/issues/12.

amchang:
Hello, I've tried using __pdoc__["module"] = False multiple times.

For example, in the solution above I would do something like __pdoc__["package.downloaded_modules"] = False in __init__.py, however, the CLI tools seems to still traverse the "downlaoded_modules" directory which then causes a problem as not everything in that "downloaded_modules" folder is python code. 

Is there something I'm missing?

kernc:
Yeah, I'm thinking pdoc shouldn't even traverse ignored modules lest havoc. :smile: 

It should probably also follow _.gitignore_ so that pointless in-tree directories needn't even be listed in `__pdoc__`.

faerics:
I have an idea of skipping a directory containing a specially named magick file, like `.no-pdoc` or `.pdocignore`. This woukd solve all the problems of deciding whether a direacory is a package, and give a user a clear ooportunity to drop a directory from docs, even if it is a valid package.

Note that this does not give an opprtunity to exclude a file from a directory, but here we can discuss some file-based (not import-based) approach, e.g parsing the content of `.pdocignore` file. 
Generally, if a user does not need a doumentation for package / module / directory, it should not be imported.

faerics:
@kernc @poke1024 
So, is anybody working on this? I am about to take the issue, narrowing it to the points below:
1. Every package containing empty `.no-pdoc` file must be ignored with a natification to user.
2. Every package containing non-empty `.no-pdoc` file must be imported as usual, and a content of `.no-pdoc` must be saved line by line in a `Module` object.
3. Every package must skip its module if it is present in its `.pdocignore` by filepath; by full python module path; by name. 

Am I missing something?

kernc:
> like `.no-pdoc` or `.pdocignore`

What are some other _ignore_ files besides `.gitignore`?

[...]

PR #139, opened 2019-12-23
Add scanfilter to Module constructor

ghost:
One way to address https://github.com/pdoc3/pdoc/issues/99: adds a new `scanfilter` parameter that allows excluding specific submodules/sub trees.

This also moves `_iter_modules` out of the method scope, which makes it monkey-patchable for extreme cases, where one would want to override the whole package scan logic (which I need to do right now).

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -479,6 +479,23 @@ class Doc:
         return self.refname < other.refname
 
 
+def _iter_modules(paths):
+    """
+    Custom implementation of `pkgutil.iter_modules()`
+    because that one doesn't play well with namespace packages.
+    See: https://github.com/pypa/setuptools/issues/83
+    """
+    from os.path import isdir, join, splitext
+    for pth in paths:
+        for file in os.listdir(pth):
+            if file.startswith(('.', '__pycache__', '__init__.py')):
+                continue
+            if file.endswith(_SOURCE_SUFFIXES):
+                yield splitext(file)[0]
+            if isdir(join(pth, file)) and '.' not in file:
+                yield file
+
+
 class Module(Doc):
     """
     Representation of a module's documentation.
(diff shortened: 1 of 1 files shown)


kernc:
> extreme cases where one would want to override the whole package scan logic (which I need to do right now)

Can you elaborate?

kernc:
Since you're familiar with the issue, what is your opinion of the [proposed override using `__pdoc__` dict](https://github.com/pdoc3/pdoc/issues/99#issuecomment-524600606):
```py
__pdoc__ = {"some.package.module": False}  # Skips some.package.module and descendants
```

poke1024:
> > extreme cases where one would want to override the whole package scan logic (which I need to do right now)
> 
> Can you elaborate? The industry rightfully doesn't favor monkey-patching private members ...

Yes, indeed. I have a very singular use case - not at all common - where I basically build a virtual tree of packages that does not exist on the disk, i.e. I want to override the whole package tree scanning. Now, the clean way to achieve this would be to add some kind of `ModuleIterator` system that is passed into `Module`, and I started adding that. But it gets complex, and I'm probably the only person who'd ever use that. So, for the sake of simplicity, being able to monkey patch for this one very special singular use case is an advantage over forking the whole module.

poke1024:
> Since you're familiar with the issue, what is your opinion of the [proposed override using `__pdoc__` dict](https://github.com/pdoc3/pdoc/issues/99#issuecomment-524600606):
> 
> ```python
> __pdoc__ = {"some.package.module": False}  # Skips some.package.module and descendants
> ```

As far as I understand it, `__pdoc__` seems to be a way to exclude subpackages from within a package I control myself.

My use case is building docs for third-party packages (e.g. bokeh, scikit-learn) that I don't control - I need to externally exclude subpackages.

Some of these subpackages include test packages and experimental stuff with esoteric dependencies, or won't run at all due to legacy problems or hard exceptions (stuff like "this package is obsolete, don't use it", which happens with plotly). Currently pdoc will stop on the importing on these subpackages. So I need to intercept them before they even get parsed or imported.

Now, my whole use case of using pdoc on libs that already have docs might of course seem nutty. But pdoc is the best tool out there to try to build unified documentation for a set of libraries.

kernc:
> `__pdoc__` seems to be a way to exclude subpackages from within a package I control myself.

Not necessarily. As the docs state:

> The keys should be string identifiers within the scope of the module or, alternatively, _fully-qualified reference names_. [...] then key (_and its members_) will be excluded from the documentation.

So IIUC, if you set:

```py
__pdoc__ = {'sklearn': False}
```
the whole of scikit-learn should become disabled. This doesn't prevent the loud importation of odd or "broken" modules, though.

-----
> So, for the sake of simplicity, being able to monkey patch for this one very special singular use case is an advantage over forking the whole module.

If you have `scanfilter=` already, of what use is monkey patching `_iter_modules()`?

-----

> basically build a virtual tree of packages that does not exist on the disk

As an alternative approach, why not create and clean a tree of files on disk? Filesystem is a great abstraction.

PR #140, opened 2019-12-24
allow seaborn-style params like {x,y}_partial

poke1024:
This allows parameters to use curly brackets like seaborn does, e.g. `{x,y}_partial` to mean both `x_partial` and `y_partial`.

One example is here (but they are all over the place in seaborn): 

https://github.com/mwaskom/seaborn/blob/66191d8a179f1bfa42f03749bc4a07e1c0c08156/seaborn/regression.py#L514

This also turns up in their rendered docs as:

```
{x,y}_partial : strings in data or matrices
Confounding variables to regress out of the x or y variables before plotting.
```
(see https://seaborn.pydata.org/generated/seaborn.lmplot.html#seaborn.lmplot)

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -166,7 +166,9 @@ class _ToMarkdown:
                           r'(?P<desc>(?:\n(?: {4}.*|$))*)',
                           _ToMarkdown._numpy_params, body, flags=re.MULTILINE)
         else:
-            body = re.sub(r'^(?P<name>\*{0,2}\w+(?:, \*{0,2}\w+)*)'
+            name = r'(?:\w|\{\w+(?:,\w+)+\})+'
+
+            body = re.sub(r'^(?P<name>\*{0,2}' + name + r'(?:, \*{0,2}' + name + r')*)'
                           r'(?: ?: (?P<type>.*))?(?<!\.)$'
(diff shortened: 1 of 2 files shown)


pdoc/html_helpers.py
```diff
@@ -166,7 +166,9 @@ def _numpy_sections(match):
                           r'(?P<desc>(?:\n(?: {4}.*|$))*)',
                           _ToMarkdown._numpy_params, body, flags=re.MULTILINE)
         else:
-            body = re.sub(r'^(?P<name>\*{0,2}\w+(?:, \*{0,2}\w+)*)'
+            name = r'(\w|\{\w+(\s*,\s*\w+)+\})+'
```
kernc: I think this regex should be:
```py
r'(?:\w|\{\w+(?:,\w+)+\})+'
```
Naimly, anonymous groups and no extraneous whitespace. With all these doc syntax extensions fairly non-standard, I prefer to maintain them as locked down and limited as possible.

kernc:
Now that new year is over, care to finish this? :smiley:

pdoc/html_helpers.py
poke1024: Changed.

kernc:
Thank you. :beers:

PR #141, opened 2019-12-24
allow wrapped lines in "see also" numpy section

poke1024:
This is to allow multiline comments in numpy "See Also" sections.

Here's an example from the wild from seaborn:
https://github.com/mwaskom/seaborn/blob/66191d8a179f1bfa42f03749bc4a07e1c0c08156/seaborn/axisgrid.py#L2204

Currently, pdoc renders this as:

old

With this PR, this is rendered as:

new

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -150,6 +150,10 @@ class _ToMarkdown:
         return ', '.join('`{}`'.format(i) for i in simple_list.split(', '))
 
     @staticmethod
+    def _numpy_unwrap(body):
+        return re.sub(r'\n\s{4}\s*', ' ', body)  # Handle line continuation
+
+    @staticmethod
     def _numpy_sections(match):
         """
         Convert sections with parameter, return, and see also lists to Markdown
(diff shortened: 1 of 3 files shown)


pdoc/html_helpers.py
```diff
@@ -158,7 +172,7 @@ def _numpy_sections(match):
         section, body = match.groups()
         if section.title() == 'See Also':
             body = re.sub(r'^((?:\n?[\w.]* ?: .*)+)|(.*\w.*)',
-                          _ToMarkdown._numpy_seealso, body)
+                          _ToMarkdown._numpy_seealso, _ToMarkdown._numpy_unwrap(body))
```
kernc: I think the unwrap method might be needlessly verbose. How about more simply:
```py
body = re.sub(r'\n\s{4}\s*', ' ', body)  # Handle line continuation
```

kernc:
Now that new year is over, care to finish this? :smiley:

pdoc/html_helpers.py
poke1024: Ok, that should do the trick indeed.

poke1024:
So far so good:

example

Are there any automated tests against html output?

kernc:
Applied with slight modification. Thanks!

Issue #90, opened 2019-08-05
Whitelisting using `__pdoc__` dict


Sieboldianus:
I've tested pdoc3 and found it a very nice alternative to Sphinx. However, my package doesn't seem to be fully documented. I've a main module which can be imported, and a [\_\_main\_\_.py](https://github.com/Sieboldianus/TagMaps/blob/master/tagmaps/__main__.py) which gets activate if a user directly runs my package in a CLI, e.g.:

```python
if __name__ == "__main__":
    main()
```

This is a pretty standard Python layout. I am not sure if I this is expected behavior for pdoc3.

### Expected Behavior

[\_\_main\_\_.py](https://github.com/Sieboldianus/TagMaps/blob/master/tagmaps/__main__.py) should be automatically documented.

### Actual Behavior

[\_\_main\_\_.py](https://github.com/Sieboldianus/TagMaps/blob/master/tagmaps/__main__.py) isn't included in pdoc3 generated docs

### Steps to Reproduce

1. Clone this [repo](https://github.com/Sieboldianus/TagMaps)
2. Install locally without deps `python setup.py --no-deps`
3. run `pdoc --html tagmaps`

### Additional info

- pdoc version: 0.3.11

kernc:
I believe any modules starting with an underscore are treated as private API.
https://github.com/pdoc3/pdoc/blob/a4a9a22056e8c2182ff5ea8d65b4718ee79439bc/pdoc/__init__.py#L597-L599

IMHO it'd be best advised to document \_\_main\_\_.py, which really only acts as main entry point into your package when run as `python -m your_package`, within \_\_init\_\_.py. Does that make sense? :thinking:

Certainly I don't want to make any dunder module always listed as such.

I'd welcome a feature to be able to include specific identifiers with:
```py
__pdoc__[some_identifier] = True
__pdoc__['your_project.__main__'] = True
```
Similarly to how they can be forced [excluded with `False`](https://pdoc3.github.io/pdoc/doc/pdoc/#overriding-docstrings-with-__pdoc__).

Sieboldianus:
> IMHO it'd be best advised to document __main__.py, which really only acts as main entry point into your package when run as `python -m your_package`, within __init__.py. Does that make sense? 🤔

Yes, that's how it works for most packages. __main__.py is really a [special convention](https://docs.python.org/3/library/__main__.html) by python, which would justify special treatment by pdoc3, e.g. `__pdoc__['your_project.__main__'] = True` 👍

[...]

PR #144, opened 2020-01-10
Allow whitelisting using __pdoc__

jpoeppel:
This PR implements whitelisting functionality for pdoc3 (#90). If the `__pdoc__` dictionary on a module level now contains a function or class name with the boolean value True, this function/class will now also be documented, even if it is otherwise marked by as private, i.e. by starting with an '_'. Also added a testcase for this behavior.

Since I just took a quick shot at it because I run into the case that I wanted to document a `__call__` override for one of my classes, I am not sure if I missed something. 

So far I assume that the whitelisting will always be done on the module level, i.e. the `__pdoc__` should be defined there instead of inside the class/function etc and just contain the full name, e.g. `A.__call__`. I think keeping all whitelisting together in one place is better than having them spread across all classes of a module, but let me know if you disagree.

One thing that I did not add, was "private" Parameters of functions, primarily since this would have required me to change the `_params` static function in order to get the function's name there, but also because I feel like specifying these parameters may be tedious.

Happy to get your comments on this.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -279,10 +279,9 @@ def _is_whitelisted(ident_name, module):
     Returns `True` if `ident_name` is contained in the module's __pdoc__ 
     with a value of `True`.
     """
-    if module:
-        pdoc = getattr(module, "__pdoc__", {})
-        if ident_name in pdoc and pdoc[ident_name] is True:
-            return True
+    pdoc = getattr(module, "__pdoc__", {})
+    if ident_name in pdoc and pdoc[ident_name]:
+        return True
 
 def _is_public(ident_name):
     """
@@ -661,6 +660,8 @@ class Module(Doc):
                         del self._context[key]
 
                 continue
+
+            # Prevent the ValueError below in case of whitelisting with "True"
             if docstring is True:
                 continue
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -267,13 +267,8 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
         else:
             continue
 
-        if not _is_public(name):
-            if isinstance(doc_obj, Class):
-                w_name = ".".join((doc_obj.name, name))
-            else:
-                w_name = name
-            if not _is_whitelisted(w_name, doc_obj.module):
-                continue
+        if not _is_public(name) and not _is_whitelisted(name, doc_obj):
+            continue
 
         docstring = inspect.cleandoc(str_node.value.s).strip()
         if not docstring:
@@ -284,25 +279,18 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
     return vars, instance_vars
 
 
-def _is_whitelisted(ident_name, doc_module):
+def _is_whitelisted(name: str, doc_obj: Union['Module', 'Class']):
     """
-    Returns `True` if `ident_name` is contained in the module's __pdoc__
-    with a value of `True`.
+    Returns `True` if `name` (relative or absolute refname) is
+    contained in some module's __pdoc__ with a truish value.
     """
-    # Check this objects module's __pdoc__ for both the relative and
-    # full canonical name
-    module_basename = doc_module.name.split(".")[-1]
-    pdoc = getattr(doc_module.obj, "__pdoc__", {})
-    if pdoc.get(ident_name, False) or \
-            pdoc.get("{}.{}".format(module_basename, ident_name), False):
-        return True
-
-    # We also need to check all potential supermodules' __pdoc__s
-    # due to full reference
-    if doc_module.supermodule:
-        new_ident = "{}.{}".format(module_basename, ident_name)
-        return _is_whitelisted(new_ident, doc_module.supermodule)
-
+    refname = doc_obj.refname + '.' + name
+    module = doc_obj.module
+    while module:
+        qualname = refname[len(module.refname) + 1:]
+        if module.__pdoc__.get(qualname) or module.__pdoc__.get(refname):
+            return True
+        module = module.supermodule
     return False
 
 
@@ -578,8 +566,7 @@ class Module(Doc):
 
             public_objs = [(name, inspect.unwrap(obj))
                            for name, obj in inspect.getmembers(self.obj)
-                           if ((_is_public(name) or
-                                _is_whitelisted(name, self.module)) and
+                           if ((_is_public(name) or _is_whitelisted(name, self)) and
                                (is_from_this_module(obj) or name in var_docstrings))]
             index = list(self.obj.__dict__).index
             public_objs.sort(key=lambda i: index(i[0]))
@@ -618,7 +605,7 @@ class Module(Doc):
                     continue
 
                 # Ignore if it isn't exported
-                if not _is_public(root) and not _is_whitelisted(root, self.module):
+                if not _is_public(root) and not _is_whitelisted(root, self):
                     continue
 
                 assert self.refname == self.name
@@ -645,6 +632,11 @@ class Module(Doc):
                 self._context.update((obj.refname, obj)
                                      for obj in docobj.doc.values())
 
+    @property
+    def __pdoc__(self):
+        """This module's __pdoc__ dict, or an empty dict if none."""
+        return getattr(self.obj, '__pdoc__', {})
+
     def _link_inheritance(self):
         # Inherited members are already in place since
         # `Class._fill_inheritance()` has been called from
@@ -657,7 +649,12 @@ class Module(Doc):
             # errors if `pdoc.link_inheritance()` is called multiple times.
             return
 
-        for name, docstring in getattr(self.obj, "__pdoc__", {}).items():
+        # Apply __pdoc__ overrides
+        for name, docstring in self.__pdoc__.items():
+            # In case of whitelisting with "True", there's nothing to do
+            if docstring is True:
+                continue
+
             refname = "%s.%s" % (self.refname, name)
             if docstring in (False, None):
                 if docstring is None:
@@ -682,10 +679,6 @@ class Module(Doc):
                     if key.startswith(refname + '.'):
                         del self._context[key]
 
-                continue
-
-            # Prevent the ValueError below in case of whitelisting with "True"
-            if docstring is True:
                 continue
 
             dobj = self.find_ident(refname)
@@ -836,9 +829,8 @@ class Class(Doc):
                        for _name, obj in inspect.getmembers(self.obj)
                        # Filter only *own* members. The rest are inherited
                        # in Class._fill_inheritance()
-                       if _name in self.obj.__dict__ and
-                       (_is_public(_name) or _is_whitelisted("{}.{}".format(
-                           name, _name), self.module))]
+                       if _name in self.obj.__dict__
+                       and (_is_public(_name) or _is_whitelisted(_name, self))]
         index = list(self.obj.__dict__).index
         public_objs.sort(key=lambda i: index(i[0]))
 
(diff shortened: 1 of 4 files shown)


pdoc/test/__init__.py
```diff
@@ -513,6 +513,28 @@ def test__pdoc__invalid_value(self):
...
+            pdoc.link_inheritance()
+            self.assertIn('A', mod.doc)
+            self.assertIn('__call__', mod.doc['A'].doc)
+
+        with patch.object(module, '__pdoc__', {'_private_function': True}):
+            mod = pdoc.Module(module)
+            pdoc.link_inheritance()
+            self.assertIn('_private_function', mod.doc)
+
+
```
kernc: Please add two more tests:
```py
# Defined in example_pkg, referring to a member of its submodule 
with patch.object(module, '__pdoc__', {'subpkg.A.__call__': True}):
```
and
```py
# Using full refname
with patch.object(module, '__pdoc__', {'example_pkg.subpkg.A.__call__': True}):
```

pdoc/test/example_pkg/__init__.py
```diff
@@ -24,6 +24,11 @@ def object_as_arg_default(*args, a=object(), **kwargs):
     """Html-encodes angle brackets in params"""
 
 
+def _private_function():
+    """Private function, should only appear if whitelisted"""
+
+
+
```
kernc: [CI is failing](https://travis-ci.org/pdoc3/pdoc/jobs/635340060#L241-L263) due to this extra whitespace.

pdoc/__init__.py
```diff
@@ -274,6 +274,15 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
 
     return vars, instance_vars
 
+def _is_whitelisted(ident_name, module):
+    """
+    Returns `True` if `ident_name` is contained in the module's __pdoc__ 
+    with a value of `True`.
+    """
+    if module:
```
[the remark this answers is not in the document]
jpoeppel: You are right, this is not necessary anymore. When I was still considering allowing it for params as well, it may be None as the _params method has the module only as optional parameter.

pdoc/__init__.py
```diff
@@ -650,6 +661,8 @@ def _link_inheritance(self):
                         del self._context[key]
 
                 continue
+            if docstring is True:
```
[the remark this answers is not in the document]
jpoeppel: I inserted it to prevent the ValueError that assumes it to be a string at this point, but I've added a comment describing this.

pdoc/test/__init__.py
jpoeppel: I am currently struggeling with these two tests, as I do not know what the best way would be to pass down the __pdoc__ of example_pkg down to subpkg. So far from my tests, it seems like subpkg will have it's own __pdoc__. I can check for supermodules and look there, but this would result in an ugly mess of special checks. The cleaner solution would probably be to update a subpackages/submodule's __pdoc__ with the supermodule's __pdoc__ or at least with the relevant elements. Am I missing some way of doing this already?

pdoc/test/example_pkg/__init__.py
jpoeppel: Should be fixed now.

pdoc/test/__init__.py
kernc: Just as a quick idea, before I even view the code, we have `pdoc.Context`.
https://github.com/pdoc3/pdoc/blob/196c93bbf0964f971ab0e2f323f9fac29fe7a7f8/pdoc/__init__.py#L63-L75
Maybe something can be saved in there for the children modules in advance?

pdoc/test/__init__.py
```diff
@@ -513,6 +513,48 @@ def test__pdoc__invalid_value(self):
...
+            self.assertIn('__call__', mod.doc['subpkg'].doc['A'].doc)
+
+        # Using full refname
+        with patch.object(module, '__pdoc__', {'example_pkg.subpkg.A.__call__': True}):
+            mod = pdoc.Module(module)
+            pdoc.link_inheritance()
+            self.assertIn('subpkg', mod.doc)
+            self.assertIn('A', mod.doc['subpkg'].doc)
+            self.assertIn('__call__', mod.doc['subpkg'].doc['A'].doc)
+
```
kernc: Let's also test whitelisting modules:
```py
with patch.object(module, '__pdoc__', {'example_pkg._private': True}):
    ...
with patch.object(module, '__pdoc__', {'_private': True}):
    ...
```

pdoc/test/__init__.py
kernc: `pdoc.Context` certainly is how the core `pdoc.Module.find_ident()` works. I think this might be a clearer approach ...

kernc:
Hmmm, I assumed docstrings can be overridden using canonical refnames in parent modules. Looking at the code, this appears to not be the case, only _current_ `__pdoc__` is iterated through. :thinking: 
https://github.com/pdoc3/pdoc/blob/196c93bbf0964f971ab0e2f323f9fac29fe7a7f8/pdoc/__init__.py#L627-L660

pdoc/test/__init__.py
jpoeppel: Do you suggest filling Module._context with the conents of __pdoc__ or rather inserting the dictionary __pdoc__ in the context so that it is available for submodules? 
The first option could lead to unintended behavior, as find_ident is used to get a Doc-Object but __pdoc__ would be having booleans or Strings as values which may even overwrite existing elements.
The second would make the assumption that there should never be anything named "__pdoc__", which might be ok.

pdoc/test/__init__.py
jpoeppel: Those tests worked out of the box with the current implementation 👍

jpoeppel:
I guess the amiguousity is only really a problem, if a subpackage has the same name as its containing package and both packages contain an object (class or function) with the same name. In that case, a specification without the global refname could mean either (both for whitelisting and overwriting), right? 
Otherwise, you would know whether or not the string represents a relative or global reference. 

In terms of current behavior for the whitelisting implementation:
In case we have: 
package
  - _A
  - package
    - _A

and we whitelist using the relative reference`package._A: True` on the top level, we would whitelist both "_A" classes/functions, however if we were to use the global reference name `package.package._A: True` on top level, it should work as intended to only whitelist the lower one. However, if we only wanted to whitelist the top level `_A`, we would need to use the relative name `_A: True`.

pdoc/test/__init__.py
kernc: Context should certainly remain mapping only refnames to doc objects. As a third option, we could attach an extra property on it, like `Context.__pdoc__`, a conjunction of all `__pdoc__` dicts we came across so far.

kernc:
But, then, if we used just `_A: True`, that would whitelist any `_A` in any later module ...

jpoeppel:
> But, then, if we used just `_A: True`, that would whitelist any `_A` in any later module ...

This is not true with the current implementation, as submodules would look for "submodule._A" in the __pdoc__ of its supermodule. I've added a test for that with the _private_function.

pdoc/test/__init__.py
jpoeppel: Not sure if this would be cleaner, as we then run into the problem discussed below with relative vs absolute references. You would need to either filter out references for the current module, or at least modify them so that submodules can clearly tell that those names are not relevant for them.

pdoc/__init__.py
```diff
@@ -275,6 +275,28 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
     return vars, instance_vars
 
 
+def _is_whitelisted(ident_name, doc_obj):
```
kernc: `doc_obj` seems only used for `doc_obj.module`, so why not accept simply `doc_module` where whitelisted `ident_name` is sought in?

pdoc/__init__.py
```diff
@@ -275,6 +275,28 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+    # full canonical name
+    module_basename = doc_obj.module.name.split(".")[-1]
+    pdoc = getattr(doc_obj.module.obj, "__pdoc__", {})
+    if pdoc.get(ident_name, False) or \
+            pdoc.get("{}.{}".format(module_basename, ident_name), False):
+        return True
+
+    # We also need to check all potential supermodules' __pdoc__s
+    # due to full reference
+    if hasattr(doc_obj.module, "supermodule") and doc_obj.module.supermodule:
```
kernc: `pdoc.Module.supermodule` is always set?
https://github.com/pdoc3/pdoc/blob/72bd9821e6a8180230f220a453dd4c0d53832063/pdoc/__init__.py#L522-L525

jpoeppel: This was also introduced when I was considering different kinds of objects (I was initially checking if "supermodule" was in doc_obj directly, but this has also been reverted. I still need the if check, because it may be None as far as I can tell.

pdoc/__init__.py
jpoeppel: You are right, I changed this when I was figuring out a way to handle the absolute paths, but ended up only needing the module after all, changed this back in newest commit.

jpoeppel:
I did add a test, using a string instead of "True" in the overwrite. The test checks if the object is whitelisted, however I am not sure how to test, if the overwrite works correctly, with regard to the resulting documentation. It should, as this should happen as before after the docs have been generated, all I am changing is the list of objects to create the doc for, all post processing should work as before.

pdoc/__init__.py
kernc: Tidying up afterwards is part of the deal. :stuck_out_tongue:

jpoeppel: True, thanks for spotting!

pdoc/test/__init__.py
```diff
@@ -513,6 +513,61 @@ def test__pdoc__invalid_value(self):
...
+        self.assertIn('A', mod.doc)
+        self.assertNotIn('__call__', mod.doc['A'].doc)
+        self.assertNotIn('_private_function', mod.doc)
+
+        with patch.object(module, '__pdoc__', {'A.__call__': "Overwrite private function doc"}):
+            mod = pdoc.Module(module)
+            pdoc.link_inheritance()
+            self.assertIn('A', mod.doc)
+            self.assertIn('__call__', mod.doc['A'].doc)
+
```
kernc: I think `mod.doc['A'].doc['__call__'].docstring` should be the expected string.

jpoeppel: Looks good, test passes when I check for that.

jpoeppel:
Any idea why Travis is listed as pending? When I check details it shows up as passed.

jpoeppel:
Is there anything still missing for this to be merged? The "problem" with relative vs global references is a general one I suppose, not something limited to this PR or am I missunderstanding something?

kernc:
The concern that popped up while fixing https://github.com/pdoc3/pdoc/commit/e8821eee42de2eaa4b56740a4ff1350f3902b354 was whether whitelisting also works for [instance variables](https://pdoc3.github.io/pdoc/doc/pdoc/#docstrings-for-variables).

That's not what's been stalling the PR, though. I should certainly get to in a few day. Thanks. :sweat_smile:

jpoeppel:
Good to know. I rebased and added a test for that. It obviously did not work right from the start, but when I check for the whitelist in the pep224 function, it does pass the tests. 
I do however need to do it a little uglier than at other places, because I need to check if doc_obj is a class or a module and prepaid the class' name in that case, because the whitelist would need to be for B._private_instance_var in order to be unique. 
Let me know if you do not like this approach and I revert that change.

kernc:
Good thing I take so long with this one. :sweat_smile: I noticed the documentation mention of the new/extended feature is still missing ... Understandably, I reserve the right to a final edit, but maybe you'd like to give copy a shot?

jpoeppel:
I updated the documentation.md file in the section regarding the overriding. Let me know what you think.

kernc:
Sorry I took so long with this one. It's an important feature, making the time well invested. Thanks for your work on this!

PR #148, opened 2020-01-17
Generate better docs when using pybind11

knarfS:
This two commits will generate nicer looking docs when documenting modules generated by pybind11 with no source code available:

* Use the `__pdoc__` Dict for documenting class variables.
* Parse the signature string in the first line of the docstring when available.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1214,12 +1214,21 @@ class Function(Doc):
                                             self.docstring, re.MULTILINE),
                                  key=len, reverse=True)
                 if strings:
-                    my_locals, my_globals = {}, self.module.obj.__dict__
+                    string = filter(strings[0])
+                    _locals, _globals = {}, {}
+                    _globals.update(typing.__dict__)
+                    _globals.update(self.module.obj.__dict__)
+                    # Trim binding module basename from type annotations
+                    # See: https://github.com/pdoc3/pdoc/pull/148#discussion_r407114141
+                    module_basename = self.module.name.rsplit('.', maxsplit=1)[-1]
+                    if module_basename in string and module_basename not in _globals:
+                        string = re.sub(r'(?<!\.)\b{}\.\b'.format(module_basename), '', string)
+
                     try:
-                        exec('def {}: pass'.format(filter(strings[0])), my_globals, my_locals)
+                        exec('def {}: pass'.format(string), _globals, _locals)
                     except SyntaxError:
                         continue
-                    signature = inspect.signature(my_locals[self.name])
+                    signature = inspect.signature(_locals[self.name])
                     if cleanup_docstring and len(strings) == 1:
                         # Remove signature from docstring variable
                         self.docstring = self.docstring.replace(strings[0], '')
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -336,23 +336,6 @@ def _toposort(graph: Mapping[T, Set[T]]) -> Generator[T, None, None]:
     assert not graph, "A cyclic dependency exists amongst %r" % graph
 
 
-def _return_annotation(name, module, obj, link=None):
-    try:
-        annot = typing.get_type_hints(obj).get('return', '')
-    except NameError as e:
-        warn("Error handling return annotation for {}: {}".format(name, e.args[0]))
-        annot = inspect.signature(inspect.unwrap(obj)).return_annotation
-        if annot == inspect.Parameter.empty:
-            annot = ''
-    if not annot:
-        return ''
-    s = inspect.formatannotation(annot).replace(' ', '\N{NBSP}')  # Better line breaks
-    if link:
-        from pdoc.html_helpers import _linkify
-        s = re.sub(r'[\w\.]+', partial(_linkify, link=link, module=module), s)
-    return s
-
-
 def link_inheritance(context: Context = None):
     """
     Link inheritance relationsships between `pdoc.Class` objects
@@ -819,7 +802,7 @@ class Class(Doc):
     """
     Representation of a class' documentation.
     """
-    __slots__ = ('doc', '_super_members')
+    __slots__ = ('doc', '_super_members', '_cached_docstring_signature')
 
     def __init__(self, name, module, obj, *, docstring=None):
         assert inspect.isclass(obj)
@@ -834,8 +817,6 @@ class Class(Doc):
 
         self.doc = {}  # type: Dict[str, Union[Function, Variable]]
         """A mapping from identifier name to a `pdoc.Doc` objects."""
-
-        pdoc = getattr(obj, '__pdoc__', {})
 
         public_objs = [(name, inspect.unwrap(obj))
                        for name, obj in inspect.getmembers(self.obj)
@@ -854,25 +835,10 @@ class Class(Doc):
                     name, self.module, obj, cls=self,
                     method=not self._method_type(self.obj, name))
             else:
-                pdoc_string = None
-
-                # Check for short refname
-                refname = self.qualname + '.' + name
-                if refname in pdoc.keys():
-                    pdoc_string = pdoc[refname]
-                    if pdoc_string is False:
-                        continue
-                # Check for full refname
-                refname = self.module.name + '.' + refname
-                if refname in pdoc.keys() and not pdoc_string:
-                    pdoc_string = pdoc[refname]
-                    if pdoc_string is False:
-                        continue
-
                 self.doc[name] = Variable(
                     name, self.module,
                     docstring=(
-                        pdoc_string or var_docstrings.get(name) or
+                        var_docstrings.get(name) or
                         (inspect.isclass(obj) or _is_descriptor(obj)) and inspect.getdoc(obj)),
                     cls=self,
                     obj=getattr(obj, 'fget', getattr(obj, '__get__', None)),
@@ -1063,7 +1029,7 @@ class Function(Doc):
     """
     Representation of documentation for a function or method.
     """
-    __slots__ = ('cls', 'method')
+    __slots__ = ('cls', 'method', '_cached_docstring_signature')
 
     def __init__(self, name, module, obj, *, cls: Class = None, method=False):
         """
@@ -1117,25 +1083,18 @@ class Function(Doc):
     def return_annotation(self, *, link=None):
         """Formatted function return type annotation or empty string if none."""
         try:
-            annot = typing.get_type_hints(self.obj).get('return', '')
+            annot = typing.get_type_hints(self.obj)['return']
         except NameError as e:
-            warn("Error handling return annotation for {}: {}".format(self.name, e.args[0]))
-            annot = inspect.signature(inspect.unwrap(self.obj)).return_annotation
-            if annot == inspect.Parameter.empty:
+            warn("Error handling return annotation for {}: {}".format(self.refname, e.args[0]))
+            annot = inspect.signature(self.obj).return_annotation
+        except (KeyError, TypeError):
+            try:
+                # Extract annotation from the docstring for C builtin function
+                annot = Function._signature_from_string(self).return_annotation
+            except AttributeError:
                 annot = ''
-        except TypeError:
-            # Extract annotation from the first line of the docstring for C builtin function
-            f_pattern = re.compile(r'^[a-zA-Z_]\w*\(.*\)'
-                                   r'(?: -> (?P<return_anno>.+))?$\n', re.MULTILINE)
-            f_match = f_pattern.match(self.obj.__doc__)
-            if f_match is None:
-                annot = ''
-            else:
-                annot = f_match.group("return_anno")
-                # Remove signature from docstring variable
-                self.docstring = f_pattern.sub('', self.docstring)
 
-        if not annot:
+        if annot is inspect.Parameter.empty or not annot:
             return ''
         s = inspect.formatannotation(annot).replace(' ', '\N{NBSP}')  # Better line breaks
         if link:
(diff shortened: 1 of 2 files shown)


knarfS:
Heck, I have no idea why these tests are failing....

knarfS:
I have some minor updates to this PR. Can I squash and force push them or do you want them as separate commits?

Any ideas about the failing tests? Any hints what's wrong?

kernc:
Commits will be squashed and/or rebased at the end, so I don't particularly mind either way.

I'm not really certain about failures. One subtest seems to be failing:
https://travis-ci.org/pdoc3/pdoc/jobs/638395055#L290
https://github.com/pdoc3/pdoc/blob/afcadcd3a4b47b4859a531f33b385d87856fb950/pdoc/test/__init__.py#L1204-L1207

The other one
https://travis-ci.org/pdoc3/pdoc/jobs/638395056#L242
is related to typing, but I can't even find the supposed affected line. :open_mouth:

knarfS:
I've done some investigation on the failing tests:

- Job #403.1 and #403.2:
I don't even know what `/csv.ext` is or what those test are doing...

- Job #403.3:
I guess that is referring to line 1232 (not 1237):
https://github.com/pdoc3/pdoc/blob/6885ab530fe03e975730b556a6bf869dd5717026/pdoc/__init__.py#L1231-L1232
Haven't changed that in my commits....

@kernc: Any help on those failing tests is appreciated!

knarfS:
Finally all tests have passed!
Also rebased to current master.

pdoc/__init__.py
```diff
@@ -1098,24 +1127,59 @@ def params(self, *, annotate: bool = False, link: Callable[[Doc], str] = None) -
...
+    def _params(doc_obj, annotate=False, link=None, module=None):
         try:
-            signature = inspect.signature(inspect.unwrap(func_obj))
+            signature = inspect.signature(inspect.unwrap(doc_obj.obj))
         except ValueError:
-            # I guess this is for C builtin functions?
-            # TODO: Extract signature from the first line of the docstring, i.e.
-            # https://github.com/mitmproxy/pdoc/commit/010d996003bc5b72fcf5fa515edbcc0142819919
-            return ["..."]
+            if doc_obj.obj.__doc__ is None:
```
kernc: Let's make this a separate function/staticmethod `_signature_from_docstring()`. Use cleaned `doc_obj.docstring` instead of `__doc__`?

pdoc/__init__.py
```diff
@@ -1098,24 +1127,59 @@ def params(self, *, annotate: bool = False, link: Callable[[Doc], str] = None) -
...
+                return ["..."]
+
+            # Extract signature from the first line of the docstring for C builtin function
```
kernc: Can you show some examples how the signature in docstring looks "in the wild"? It makes regexes easier to follow. Where in CPython is that stringified signature constructed?

pdoc/__init__.py
```diff
@@ -1098,24 +1127,59 @@ def params(self, *, annotate: bool = False, link: Callable[[Doc], str] = None) -
...
         """
-        return self._params(self.obj, annotate=annotate, link=link, module=self.module)
+        return self._params(self, annotate=annotate, link=link, module=self.module)
 
     @staticmethod
-    def _params(func_obj, annotate=False, link=None, module=None):
+    def _params(doc_obj, annotate=False, link=None, module=None):
         try:
-            signature = inspect.signature(inspect.unwrap(func_obj))
+            signature = inspect.signature(inspect.unwrap(doc_obj.obj))
```
kernc: Since Python 3.5, `inspect.unwrap()` seems to be [not needed](https://docs.python.org/3.5/library/inspect.html#inspect.signature).

pdoc/__init__.py
kernc: So, if the string signatures are typically valid Python code, I think it might be more manageable and far less intricate to reach signature object with something dumb like:
```py
my_locals = {}
exec('def ' + string_signature + ': pass', {}, my_locals)
inspect.signature(my_locals[doc_obj.name])
```

pdoc/__init__.py
```diff
@@ -826,10 +828,25 @@ def __init__(self, name, module, obj, *, docstring=None):
...
+                if refname in pdoc.keys():
+                    pdoc_string = pdoc[refname]
+                    if pdoc_string is False:
+                        continue
+                # Check for full refname
+                refname = self.module.name + '.' + refname
+                if refname in pdoc.keys() and not pdoc_string:
+                    pdoc_string = pdoc[refname]
+                    if pdoc_string is False:
+                        continue
```
kernc: What are these two blocks? :eyes:

pdoc/__init__.py
knarfS: > Can you show some examples how the signature in docstring looks "in the wild"? It makes regexes easier to follow.

The signatures have to be valid function definitions (without the `def` keyword) and can be as simple as
```
foo(bar)
```
or fully fletched with type hints, etc.:
```
foo(bar1: str, bar2: int = 1) -> Tuple[int, str]
```
Here is a real world example from my `smuview` module. The signature was generated by pybind11:
```
add_user_channel(self: smuview.BaseDevice, channel_name: str, channel_group_name: str) -> smuview.UserChannel
```
> Where in CPython is that stringified signature constructed?

I doubt (but I don't know) if there is a CPython function, which generates the function signature. In my case this is done by [pybind11](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx) itself:
> By default, pybind11 automatically generates and prepends a signature to the docstring of a function [...]
  
The stringified signature is documented in PEP 257, [Capter "One-line Docstrings"](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings):
> The one-line docstring should NOT be a "signature" reiterating the function/method parameters (which can be obtained by introspection).
> [...]
> This type of docstring is only appropriate for C functions (such as built-ins), where introspection is not possible.
  
And from the [numpy docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html#sections):
> The function signature is normally found by introspection and displayed by the help function. For some functions (notably those written in C) the signature is not available, so we have to specify it as the first line of the docstring: [...]

knarfS: > So, if the string signatures are typically valid Python code, I think it might be more manageable and far less intricate to reach signature object with something dumb like:
> 
> ```python
> my_locals = {}
> exec('def ' + string_signature + ': pass', {}, my_locals)
> inspect.signature(my_locals[doc_obj.name])
> ```

Great! A first test worked!
Much simpler, but you have to use the actual `locales()` to get the imported stuff of the module that should be documented. I don't know if this conflicts with something else, but I will test.

pdoc/__init__.py
knarfS: This is for retrieving the docstring of class variables from the `__pdoc__` Dict.

First lookup (block) is for the short reference name (e.g.: `class_name.variable_name`) the second is for the full reference name (`module_name.class_name.variable_name`), as one or the other could be used. When set to `False`, the variable will be excluded from the documentation, as defined [here](https://pdoc3.github.io/pdoc/doc/pdoc/#overriding-docstrings-with-__pdoc__).

pdoc/__init__.py
knarfS: I will change this.

pdoc/__init__.py
knarfS: I will test and change this.

pdoc/__init__.py
kernc: I don't get it. Are you saying that without this addition the described documented behavior doesn't, in fact, work? :open_mouth:

knarfS: Yes you are correct: Entries for class variables in the `__pdoc__ ` Dict are not evaluated at the moment.

kernc: Yikes. :sweat_smile: Would you open a separate PR for that?

knarfS: Sure! Will do later today...

Should I remove commit f77f882 and rebase this PR? Will this conversation be lost when I rebase the PR?

kernc: I think it gets obsoleted/collapsed once the touching lines are changed.

kernc: Continued in https://github.com/pdoc3/pdoc/pull/162.

pdoc/__init__.py
kernc: Did you then manage to push this further? :smiley:

knarfS: Yes, I've done more testing with `exec()` and `inspect.signature()` and run into a problem.
When you have python built-in types as type hints, this is working great. But when the signature contains types from other packages, it starts to get complicated.

pdoc/__init__.py
```diff
@@ -1196,6 +1212,36 @@ def __repr__(self):
...
+                    # (e.g. `range` or `slice` or `itertools.repeat` or `numpy.arange`)
+                    (r'^{}\(.*\)(?= -|$)', False, lambda s: s.replace('[', '').replace(']', '')),
+            ):
+                strings = sorted(re.findall(expr.format(self.name),
+                                            self.docstring, re.MULTILINE),
+                                 key=len, reverse=True)
+                if strings:
+                    my_locals, my_globals = {}, self.module.obj.__dict__
+                    try:
+                        exec('def {}: pass'.format(filter(strings[0])), my_globals, my_locals)
```
kernc, line 1234: > https://github.com/pdoc3/pdoc/pull/148#discussion_r407103658

Updated proposal includes globals from wherever the object is defined. Would this work?

pdoc/__init__.py
```diff
@@ -1196,6 +1212,36 @@ def __repr__(self):
...
 
+    @staticmethod
+    def _signature_from_string(self):
+        try:
+            return self._cached_docstring_signature
+        except AttributeError:
+            signature = None
+            for expr, cleanup_docstring, filter in (
+                    # Full proper typed signature, such as one from pybind11
+                    (r'^{}\(.*\)(?: -> .*)?$', True, lambda s: s),
```
kernc, line 1223: Will this RE work with pybind11? The [example in their documentation](https://pybind11.readthedocs.io/en/master/basics.html#default-arguments) indicates the signature docstring line looks like this?
```
Signature : (i: int = 1, j: int = 2) -> int

A function which adds two numbers ...
```

knarfS: No, the signature docstring for that specific function in the example would be:
```
add(i: int = 1, j: int = 2) -> int
```

The example is a formated output (probably from pydoc.help()).

pdoc/__init__.py
knarfS: I ran into the same issue I mentioned above.
For example the signature
```
get_sample(self: smuview.AnalogSampleSignal, pos: int) -> Tuple[int, float]
```
will cause the error
```
NameError: name 'smuview' is not defined
```

Because `AnalogSampleSignal` is in `self.module.obj.__dict__`, but not `smuview.AnalogSampleSignal`, I've changed the signature to
```
get_sample(self: AnalogSampleSignal, pos: int) -> Tuple[int, float]
```
But then I get the error
```
NameError: name 'Tuple' is not defined
```

Adding `from typing import Tuple` to the exec string works until the next unknown type occurs.

pdoc/__init__.py
kernc: I don't think `help` reformats. Must be an old doc example then. Thanks.

pdoc/__init__.py
kernc: > I've changed the signature to

How did you do that? Does pybind11 generally refer to its exported members with a module prefix?

We can prepopulate passed globals with `typing` members. We can't catch them all cases, but sure can try a few. :D

knarfS: Also ast (no type hints) and typed_ast (only type comments) are useless.
Only solutions I see atm is a regex parser (from my earlier commit) with the risk of not catching complex types. Or a new sophisticated (classic) parser for signatures.

knarfS: > > I've changed the signature to
> 
> How did you do that?

Hacked into `_signature_from_string()` with `.replace('smuview.', '')` for quick'n'dirty testing ;)

> Does pybind11 generally refer to its exported members with a module prefix?

Yes, afaict.

knarfS: Here is the documentation for the whole smuview module, generated with `pydoc.help(smuview)`. The docstrings are pretty much unformated: 
https://github.com/knarfS/smuview/blob/master/doc/smuview_python_bindings.txt

Here is the documentation generated by pdoc with my patches applied:
https://github.com/knarfS/smuview/blob/master/doc/smuview_python_bindings.html

HTH

kernc: I think the latest pushed change should work with modules like yours.

What did you mean with ast, typed_ast?

knarfS: > What did you mean with ast, typed_ast?

The [Python Abstract Syntax Tree](https://docs.python.org/3/library/ast.html) offers the possibility to parse Python code. Here is an example for function parameters: https://stackoverflow.com/a/51271704

Since version 3.8 ast supports [type comments](https://www.python.org/dev/peps/pep-0484/#type-comments), before 3.8 there is [typed_ast](https://github.com/python/typed_ast).
But none of them is supporting type hints/annotations, so they are useless for the function signatures (definitions see [here](https://github.com/pdoc3/pdoc/pull/148#discussion_r377598769)).

knarfS:
I've added the pybind11 special datatype `capsule` to the globals dict.

knarfS:
I've added a new branch (https://github.com/knarfS/pdoc/tree/pybind11_3), that is rebased to the current master and removed the obsolete "ENH: Use __pdoc__ Dict for class variables if available." commit. But not squashed together yet.

If you want to I can move that to this PR.

All tests so far have worked, but I'm not using any 3rd party libs/types in the function signatures (yet).

kernc:
Managed to solve the merging spectacle. :sweat_smile: Let's see how far this works with real projects. Thanks for all the work on this one! :cake:

knarfS:
Thanks, everything is working fine now!

PR #153, opened 2020-01-30
BUG: Fix function parameters that shadow names from outer scope

richardchien:
This PR fixes bug in the following situation:

```python
# pkg1/__init__.py

# assume there is a submodule named "message"
from .message import Message
# after importing "Message", "message" becomes a name in global scope

def send(message: Message):
    # previously, the "message" parameter will be rendered as a link to the "message" module
    pass
```

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1172,16 +1172,19 @@ class Function(Doc):
 
             s = str(p)
             if p.annotation is not EMPTY:
+                name, annotation = s.split(':', maxsplit=1)
+                name, annotation = name.strip(), annotation.strip()
                 if sys.version_info < (3, 7):
                     # PEP8-normalize whitespace
-                    s = re.sub(r'(?<!\s)=(?!\s)', ' = ', re.sub(r':(?!\s)', ': ', s, 1), 1)
+                    annotation = re.sub(r'(?<!\s)=(?!\s)', ' = ', annotation, 1)
                 # "Eval" forward-declarations (typing string literals)
                 if isinstance(p.annotation, str):
-                    s = s.replace("'%s'" % p.annotation, p.annotation, 1)
-                s = s.replace(' ', '\N{NBSP}')  # prevent improper line breaking
+                    annotation = p.annotation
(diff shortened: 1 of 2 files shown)


pdoc/__init__.py
```diff
@@ -1172,16 +1172,19 @@ def __repr__(self):
...
+                    annotation = re.sub(r'(?<!\s)=(?!\s)', ' = ', annotation, 1)
                 # "Eval" forward-declarations (typing string literals)
                 if isinstance(p.annotation, str):
-                    s = s.replace("'%s'" % p.annotation, p.annotation, 1)
-                s = s.replace(' ', '\N{NBSP}')  # prevent improper line breaking
+                    annotation = p.annotation
+                annotation = annotation.replace(' ', '\N{NBSP}')  # prevent improper line breaking
 
                 if link:
-                    s = re.sub(r'[\w\.]+', _linkify, s)
```
kernc: IIUC, the issue lies here with this linking all `\w+` parts of `s`, namely the annotation _and the parameter identifier_, which should never be linked at all? :thinking: 

Sorry, I find your solution too involved. :sweat_smile: Think of something simpler or let me think it through.

richardchien: > IIUC, the issue lies here with this linking all `\w+` parts of `s`, namely the annotation _and the parameter identifier_

Yes. 

IMO, the _parameter identifier_ (in the `inspect.Parameter` object, it's called _name_) should not be linked, because it is syntactically legal to name a parameter the same as another object in outer scope, and that parameter will absolutely not refer to that outer scope object in the function scope. Also, in the case I presented in conversation, the side effect of `import` from `__init__.py` may make this even worse. OTOH, in docstrings, we can use double <code>`</code> to avoid it if we don't want those links, but in function signatures, it's impossible.

My idea in this PR is to split the _idendifier_ and the _annotation_ from `s`, and only linkify the annotation.

kernc: Indeed, _name_ shouldn't ever be linked because it's never a reference to an existing item.

There's sadly much less we can do about backtick references inside docstrings – it's simply ambiguous. :disappointed:

richardchien: Your solution is really neat. Great!👍

Issue #159, opened 2020-02-08
Scroll sidebar separately


yuehaowang:
I am a beginner of this tool. I think pdoc3 is a very convenient and powerful tool for creating API docs. But I met some problems about the sidebar. I am here to ask my questions and give my suggestions.

### Expected Behavior

1. I would like to fix the sidebar when I scroll the page down on the right content part. And scroll the sidebar down only if I scroll the page down on the left sidebar part. Here is an [example](https://threejs.org/docs/index.html#manual/en/introduction/Creating-a-scene).

2. I also think it would be better if pdoc provides an easy way to create a to-top button.

3. Is there any other free templates available online?

### Actual Behavior

1. The sidebar will be scrolled down with the scrolling of the content part.

2. I do not find out a way to create a to-top button in addition to customizing template.

3. I tried to google 'pdoc3 templates', but it seems it's hard to find something useful.

### Additional info

- pdoc version: pdoc3-0.7.4

Thanks.

kernc:
It's normally better to open separate issues for separate "questions".

1. Separate scrolling sidebar would require minor CSS adjustments. A PR would be looked upon favorably. :+1: 

2. Seems like we don't yet have a "#top" button, not even for a11y. :thinking: PR welcome.

3. I'm not aware of any template repositories. Compared to Sphinx or MkDocs, it's a fairly young and somewhat disjoint project. Also of smaller scope.
   Many styles can be easily overridden on per-needed basis (for instance, you can source a custom CSS file in [_head.mako_](https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates)).

yuehaowang:
Thanks for your response. I planned to new a PR on separate scrolling sidebar. I would like to know whether there is any contributing guide regarding code style, modification guide, etc. 

I really appreciate this project since it is very easy to get started and generates beautiful docs.

Thanks again.

kernc:
Sure, the rough [contributing guidelines](https://github.com/pdoc3/pdoc/blob/master/CONTRIBUTING.md) are here. Basically, if CI passes, that's ok. The rest is subject to popular vote.

Hope the code is easy to work with, too. :blush: The relevant sidebar bits are probably here:
https://github.com/pdoc3/pdoc/blob/1709915249cdbdf9d79f08628a1ff08190678adf/pdoc/templates/css.mako#L275-L279

PR #160, opened 2020-02-09
Support separate scrolling sidebar

yuehaowang:
Fixes https://github.com/pdoc3/pdoc/issues/159

I've edited css.mako to support the feature mentioned in #159 .

As `vw` and `vh` are used in my implementation, it may have compatibility problems on dated web browsers. I do not know whether it matters for pdoc. [Here](https://caniuse.com/#feat=viewport-units) is the supporting situation of `vw` & `vh` among different web browsers.

Follow-up discussions are welcome.


kernc:
Evergreen browsers with graceful degradation elsewhere, that's fine. Can you fix the three scrollbars?

yuehaowang:
Sorry, I did not noticed there are three scrollbars because my web browser (Google Chrome for macOS) hides scrollbars when they are not scrolling. It seems the third scrollbar is caused by the footer. I think putting the footer to the end of article part would fix this problem and may also have a good appearance. How do you think about this idea?

kernc:
I certainly don't mind it. There's usually nothing too important in there.

yuehaowang:
I've moved the footer to the article part. How about the current appearance?

kernc:
It was quite a CSS nut to crack, but this also seems to work reasonably well, and I like it better. :smiley:

PR #162, opened 2020-02-12
ENH: Use __pdoc__ Dict for class variables if available.

knarfS:
Entries for class variables in the `__pdoc__ ` Dict are not evaluated at the moment, but should be as defined [here](https://pdoc3.github.io/pdoc/doc/pdoc/#overriding-docstrings-with-__pdoc__).

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -809,6 +809,8 @@ class Class(Doc):
         self.doc = {}
         """A mapping from identifier name to a `pdoc.Doc` objects."""
 
+        pdoc = getattr(obj, '__pdoc__', {})
+
         public_objs = [(name, inspect.unwrap(obj))
                        for name, obj in inspect.getmembers(self.obj)
                        # Filter only *own* members. The rest are inherited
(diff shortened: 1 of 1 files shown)


kernc:
As this test indicates, class variables should be covered:
https://github.com/pdoc3/pdoc/blob/65b31fd01b671b0d46f78f8dd46a2eb071ddfa3d/pdoc/test/__init__.py#L503-L516

The PR asks `__pdoc__` of the _type object_ (`pdoc = getattr(obj, '__pdoc__', {})`), this premise is wrong. `__pdoc__` needs to be defined on the module, not the class.

I think the reason you experienced this issue is that you didn't define `__pdoc__` on the module beforehand. I.e., this is correct and works:
```py
__pdoc__ = {}

class Class:
    __pdoc__['Class.whatever'] = False
```

knarfS:
You are right. This is working without this patch!

For reference, here is my implementation of the `__pdoc__` attribute in pybind11: https://github.com/knarfS/pybind11/commit/4aab7474990ef66e14038e9646d313dd720f9196

kernc:
Patching up into an upstream lib like that certainly doesn't look like a way to go. I think one can export global `__pdoc__` as a [variable](https://pybind11.readthedocs.io/en/master/basics.html#exporting-variables).

PR #168, opened 2020-02-20
Fix changed command-line option of Pandoc

MPvHarmelen:


kernc:
Source? [Pandoc docs say](https://pandoc.org/MANUAL.html#option--pdf-engine) the switch is still `--pdf-engine`.

MPvHarmelen:
I remember copy-pasting the advised command and pandoc then complaining that the option `--pdf-engine` doesn't exist, but I can't reproduce it...

PR #172, opened 2020-03-04
Skip subclasses which have non-str __module__ attribute

hashbrowncipher:
I encountered a crash of pdoc on code using Pydantic's `create_model`,
which defaults to passing `__module__ = None` when creating new
subclasses. I think it's non-optimal of Pydantic to create such classes,
but I figured that it would be best for pdoc3 to be resilient to the
situation.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -897,7 +897,8 @@ class Class(Doc):
         and `pdoc.External` otherwise.
         """
         return sorted(self.module.find_class(c)
-                      for c in type.__subclasses__(self.obj))
+                      for c in type.__subclasses__(self.obj)
+                      if isinstance(c.__module__, str))
 
     def params(self, *, annotate=False, link=None) -> List['str']:
         """
--- a/pdoc/test/__init__.py
+++ b/pdoc/test/__init__.py
@@ -603,6 +603,10 @@ class ApiTest(unittest.TestCase):
         class E(C):
             pass
 
+        # H should not show up as a subclass of A, because it has anon-str __module__
+        # attribute
+        H = type("H", (A,), {"__module__": None})  # noqa: F841
+
         mod = pdoc.Module(pdoc)
         self.assertEqual([x.refname for x in pdoc.Class('A', mod, A).subclasses()],
                          [mod.find_class(C).refname])


pdoc/__init__.py
```diff
@@ -897,7 +897,8 @@ def subclasses(self) -> List['Class']:
         and `pdoc.External` otherwise.
         """
         return sorted(self.module.find_class(c)
-                      for c in type.__subclasses__(self.obj))
+                      for c in type.__subclasses__(self.obj)
+                      if isinstance(c.__module__, str))
```
kernc: It seems `__module__` can be `None` by the [Python spec](https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy). :+1:

What breakage you experienced, this doesn't feel like the right place to fix it, though. Why shouldn't subclasses without `__module__` be part of the hierarchy?

kernc:
Can you attach the full crash report?

hashbrowncipher:
> Can you attach the full crash report?

Yep.

```
  File ".tox/linters/bin/pdoc3", line 10, in <module>
    sys.exit(main())
  File ".tox/linters/lib/python3.7/site-packages/pdoc/cli.py", line 489, in main
    write_files(module, ext='.html', **template_config)
  File ".tox/linters/lib/python3.7/site-packages/pdoc/cli.py", line 329, in write_files
    write_files(submodule, ext=ext, **kwargs)
  File ".tox/linters/lib/python3.7/site-packages/pdoc/cli.py", line 329, in write_files
    write_files(submodule, ext=ext, **kwargs)
  File ".tox/linters/lib/python3.7/site-packages/pdoc/cli.py", line 318, in write_files
    w.write(m.html(**kwargs))
  File ".tox/linters/lib/python3.7/site-packages/pdoc/__init__.py", line 699, in html
    html = _render_template('/html.mako', module=self, **kwargs)
  File ".tox/linters/lib/python3.7/site-packages/pdoc/__init__.py", line 121, in _render_template
    return t.render(**config).strip()
  File ".tox/linters/lib/python3.7/site-packages/mako/template.py", line 476, in render
    return runtime._render(self, self.callable_, args, data)
  File ".tox/linters/lib/python3.7/site-packages/mako/runtime.py", line 883, in _render
    **_kwargs_for_callable(callable_, data)
  File ".tox/linters/lib/python3.7/site-packages/mako/runtime.py", line 920, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File ".tox/linters/lib/python3.7/site-packages/mako/runtime.py", line 947, in _exec_template
    callable_(context, *args, **kwargs)
  File "_html_mako", line 127, in render_body
  File "_html_mako", line 44, in show_module
  File "_html_mako", line 410, in render_show_module
  File ".tox/linters/lib/python3.7/site-packages/pdoc/__init__.py", line 900, in subclasses
    for c in type.__subclasses__(self.obj))
  File ".tox/linters/lib/python3.7/site-packages/pdoc/__init__.py", line 900, in <genexpr>
    for c in type.__subclasses__(self.obj))
  File ".tox/linters/lib/python3.7/site-packages/pdoc/__init__.py", line 729, in find_class
    return self.find_ident(cls.__module__ + '.' + cls.__qualname__)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

pdoc/__init__.py
hashbrowncipher: Yeah, there may be other approaches rather than just simply not displaying the subclass. Substituting '???' is one approach, but I imagine you might have a better idea.

kernc: Absolutely, given [the exception](https://github.com/pdoc3/pdoc/pull/172#issuecomment-595008094), I think something like replacing occurrences of `.__module__ +` with `.__module__ or '?') +` might be better? This will result in `find_ident()` _finding_ an `External("?.FooClass")`.

PR #175, opened 2020-03-25
Include annotations for class properties

postlund:
Type annotations (when enabled) are not included in the generated documentation for members in classes marked with `@property`. This is an attempt to fix that.


kernc:
Thank you. I'll try to get a release out soon.

postlund:
Awesome, thanks! :1:

PR #176, opened 2020-03-29
Add parsing of nested classes

angeloskath:
Nested classes are incorrectly parsed as class variables. This commit fixes that and adds a method `all_classes` in `Module` that returns even the nested classes.

Thanks,
Angelos

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -225,8 +225,12 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
     if _init_tree:
         tree = _init_tree
     else:
+        source = doc_obj.source
+        if source == "":
+            warn("Couldn't read PEP-224 variable docstrings from {!r}".format(doc_obj))
+            return {}, {}
         try:
-            tree = ast.parse(inspect.getsource(doc_obj.obj))
+            tree = ast.parse(source)
         except (OSError, TypeError, SyntaxError) as exc:
             warn("Couldn't read PEP-224 variable docstrings from {!r}: {}".format(doc_obj, exc))
             return {}, {}
@@ -786,6 +790,17 @@ class Module(Doc):
         """
         return self._filter_doc_objs(Class, sort)
 
+    def all_classes(self, sort=True):
+        """Returns all documented classes in the module even the ones defined
+        in other classes."""
+        stack = self.classes(sort)[::-1]
+        results = []
+        while stack:
+            c = stack.pop()
+            results.append(c)
+            stack.extend(c.classes(sort=sort)[::-1])
+        return results
+
     def functions(self, sort=True):
         """
         Returns all documented module-level functions in the module,
@@ -813,9 +828,9 @@ class Class(Doc):
     """
     Representation of a class' documentation.
     """
-    __slots__ = ('doc', '_super_members')
+    __slots__ = ('doc', 'cls', '_super_members')
 
-    def __init__(self, name, module, obj, *, docstring=None):
+    def __init__(self, name, module, obj, *, docstring=None, cls=None):
         assert inspect.isclass(obj)
 
         if docstring is None:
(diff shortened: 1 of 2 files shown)


pdoc/__init__.py
```diff
@@ -225,8 +225,12 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
     if _init_tree:
         tree = _init_tree
     else:
+        source = doc_obj.source
+        if source == "":
+            warn("Couldn't read PEP-224 variable docstrings from {!r}".format(doc_obj))
+            return {}, {}
         try:
-            tree = ast.parse(inspect.getsource(doc_obj.obj))
+            tree = ast.parse(source)
```
kernc, line 233: This makes sense, but could you merge the two into the same error. `ast.parse()` surely throws something on an empty string?

kernc (changes requested):
I'm thinking maybe we should actually make this the default (i.e. change `Module.classes()` directly). Classes are classes, not variables, and if they are exposed and meant to be used (they should probably not be nested), they better be documented thoroughly.

This also needs a unit test in `pdoc.test.__init__`.

angeloskath:
Hi, thanks for considering this.

I can make the requested changes and add the test maybe from Monday. I am just sending so that you know I haven't just dropped this here and left :-).

Cheers,
Angelos

kernc:
No hurry, but this would sure make a nice addition! :smiley:

frank101010:
Pull request https://github.com/pdoc3/pdoc/pull/390 contains most of angeloskath's changes, but based on pdoc3, commit https://github.com/pdoc3/pdoc/commit/2cce30a9b55eeeddc1ed826c8a2ada53777c3eea (March 2022).

However, for some reason I'm not able to run pdoc's unit tests as described here : https://github.com/pdoc3/pdoc/blob/2cce30a9b55eeeddc1ed826c8a2ada53777c3eea/CONTRIBUTING.md

This is what I get when running `python -m unittest -v pdoc.test`:
```
======================================================================
ERROR: test (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test
  File "C:\Projects\thirdparty\pdoc\pdoc\__init__.py", line 222, in import_module
    module = importlib.import_module(module_path)
  File "C:\Python39\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "C:\Projects\thirdparty\pdoc\pdoc\test\example_pkg\index.py", line 1, in <module>
    __init__.py
NameError: name '__init__' is not defined

  File "C:\Python39\lib\unittest\loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "C:\Projects\thirdparty\pdoc\pdoc\test\__init__.py", line 40, in <module>
    EXAMPLE_PDOC_MODULE = pdoc.Module(EXAMPLE_MODULE, context=pdoc.Context())
  File "C:\Projects\thirdparty\pdoc\pdoc\__init__.py", line 754, in __init__
    m = Module(import_module(fullname),
  File "C:\Projects\thirdparty\pdoc\pdoc\__init__.py", line 224, in import_module
    raise ImportError(f'Error importing {module!r}: {e.__class__.__name__}: {e}')
ImportError: Error importing 'example_pkg.index': NameError: name '__init__' is not defined
```
which seems due to the fact that some files in example_pkg contain nothing but a python file name.
For example the file `example_pkg/index.py` mentioned in above trace contains the single text line `__init__.py` which Python tries to execute as "access member `py` of symbol `__init__`".

Did I miss something here, perhaps some code generation step?

PR #150, opened 2020-01-18
ENH: Better parsing of markdown code spans. Fix issue #145.

knarfS:
Fixes https://github.com/pdoc3/pdoc/issues/145.

This fix works for complex types like ``Set[`MyClass`]`` and ``Tuple[`int, `MyClass`]``, but also for ``Tuple[`int`, Dict[`int`, `MyClass`], `MyClass`]``, while skipping markdown links like ``[`Text`](Url)`` and ``[`Text`]: (Url)``.

But I can't tell if that regular expression is good enough or maybe drops some of the old recognition patterns...

Issue #145, opened 2020-01-14
Links to references in type strings don't work for complex types.


knarfS:
### Expected Behavior

Pdoc3 should add links to known references (here `MyClass`) for complex types like `Set[MyClass]` or `Dict[int, MyClass]`

### Actual Behavior

Pdoc3 only adds reference in complex types when there are additional white spaces. `Set[ MyClass ]` or `Dict[ int, MyClass ]`is working.

### Additional info

In `html_helpers.py` function `to_markdown()` links between square brackets are excluded (is there a reason for this?):
```python
# Matches markdown code spans not +directly+ within links.
# E.g. `code` and [foo is `bar`]() but not [`code`](...)
# Also skips \-escaped grave quotes.
_code_refs=re.compile(r'(?<![\[\\`])`(?!])(?:[^`]|(?<=\\)`)+`').sub
```
This regular expression is working for me, but it is probably not valid for other cases:
```python
r'(?<![\\`])`(?!])(?:[^`]|(?<=\\)`)+`'
```

- pdoc version: git commit 196c93b

kernc:
References between square brackets are excluded as mentioned in the comment, because they would override what are already markdown hyperlinks:
```markdown
[`ref`](url)   # We don't want ref to overrule url
```

I think the regex might be missing an additional lookahead check for the opening URL-containing parenthesis ... :thinking:

Do the tests by chance pass with your modification?
```bash
python -m pdoc.test.__init__
```

knarfS:
Yes, the modified regular expression passes the unit tests.

kernc:
Sorry, can you give an example docstring? Even with your regex fix, I can't get
```md
`Dict[int, pdoc.Doc]`
```
to link. :flushed:

knarfS:
I've added a PR that may fix this problem.
The new regexpr should be reviewed ;)

kernc:
> This regular expression is working for me, but it is probably not valid for other cases:

Can we first have a docstring example on which your modified expression works and the original doesn't? So that the issue can be easily reproduced.

knarfS:
Test Docstrings:
```
`test`
`test` test `test`
Dict[`test`, `test`]
Tuple[`test`, `test`, `test`]
Tuple[`test`, Dict[`test`, `test`], `test`]
[`test`](URL)
[`test`]: (URL)
```

Everything between back ticks should match, except the last two (markdown links).

knarfS:
[...]

PR #181, opened 2020-04-13
BUG: Improved reference linking in convolved type annotations

kernc:
Fixes https://github.com/pdoc3/pdoc/issues/145, closes https://github.com/pdoc3/pdoc/pull/150.

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -130,40 +130,38 @@ class _ToMarkdown:
     will come in handy.
     """
     @staticmethod
-    def _deflist(name, type, desc,
-                 # Wraps any identifiers and string literals in parameter type spec
-                 # in backticks while skipping common "stopwords" such as 'or', 'of',
-                 # 'optional' ... See §4 Parameters:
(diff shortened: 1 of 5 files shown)


knarfS:
Works for SmuView (in combination with PR #148). Only minor detail is the newly false linked `pybind11_builtins.pybind11_object` ancestor.
New generated doc: https://knarfs.github.io/doc/smuview/continuous/python_bindings_api.html
Old doc: https://knarfs.github.io/doc/smuview/0.0.4/python_bindings_api.html

kernc:
Thank god for CI. That was a bit overreaching.


> the newly false linked `pybind11_builtins.pybind11_object` ancestor.


That should no longer be an issue, whereas in general, the newly linking mechanism is inherently more prone to false positives, and that's why I [limited it](https://github.com/pdoc3/pdoc/pull/181/files#diff-495df8a7d2b61993a834a295ddbfd0a7R444-R447) to type-annotation-like strings only. Thus, type annotations should hyperlink whereas arbitrary expressions like `mypkg.SomeClass + 6` won't.

knarfS:
Thanks, everything is working fine now!

PR #185, opened 2020-04-18
Add support for python 3.8 positional only argument syntax

### LINKED ISSUE #183: Support for python 3.8 positional only argument syntax.
### Expected Behavior

I'd expect the parameters for methods that use the positional only argument syntax that was introduced in python 3.8 (see [what's new to python3.8](https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters) for more information) to include `/` as to indicate that arguments to the left of `/` are positional only (e.g. ``def compile (self, method: str, /, **kwargs: Any) -> CompiledRoute``).

### Actual Behavior

The `/` argument is ignored, resulting in there being no indication that some of the arguments can only be positional, as shown in the screenshot below.
image

### Steps to Reproduce

Generate documentation for a method that uses python 3.8's positional only argument syntax (see [what's new to python3.8](https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters) for more information). The following method was used to produce the example actual behavior given above.

```python
def compile(self, method: str, /, **kwargs: typing.Any) -> CompiledRoute:
    """Generate a formatted `CompiledRoute` for this route template.

    This takes into account any URL parameters that have been passed, and extracting
    the `major_params` for bucket hash operations accordingly.

    Parameters
    ----------
    method : str
        The method to use.
    **kwargs : typing.Any
        Any parameters to interpolate into the route path.

    Returns
    -------
    CompiledRoute
        The compiled route.
    """
    major_hash_part = "-".join((str(kwargs[p]) for p in self.major_params))

    return CompiledRoute(method, self.path_template, self.path_template.format_map(kwargs), major_hash_part)
```

### Additional info

- pdoc version: 0.8.1
- python version: 3.8.2

davfsa:
Fixes #183

### DIFF
--- a/pdoc/test/__init__.py
+++ b/pdoc/test/__init__.py
@@ -802,6 +802,15 @@ class ApiTest(unittest.TestCase):
         self.assertEqual(pdoc.Function('get_sample', mod, get_sample).return_annotation(),
                          'Tuple[int,\xa0float]')
 
+    @unittest.skipIf(sys.version_info < (3, 8), "positional-only arguments unsupported in < py3.8")
+    def test_test_Function_params_python38_specific(self):
+        mod = pdoc.Module(pdoc)
+        func = pdoc.Function('f', mod, eval("lambda a, /, b: None"))
+        self.assertEqual(func.params(), ['a', '/', 'b'])
+
+        func = pdoc.Function('f', mod, eval("lambda a, /: None"))
+        self.assertEqual(func.params(), ['a', '/'])
+
     def test_Function_return_annotation(self):
         import typing
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1249,6 +1249,7 @@ class Function(Doc):
 
         params = []
         kw_only = False
+        pos_only = False
         EMPTY = inspect.Parameter.empty
 
         if link:
@@ -1258,6 +1259,12 @@ class Function(Doc):
         for p in signature.parameters.values():  # type: inspect.Parameter
             if not _is_public(p.name) and p.default is not EMPTY:
                 continue
+
+            if p.kind == p.POSITIONAL_ONLY:
+                pos_only = True
+            elif pos_only:
+                params.append("/")
+                pos_only = False
 
             if p.kind == p.VAR_POSITIONAL:
                 kw_only = True
@@ -1286,6 +1293,9 @@ class Function(Doc):
                     s = linked_annotation.join(s.rsplit(annotation, 1))  # "rreplace" once
 
             params.append(s)
+
+        if pos_only:
+            params.append("/")
 
         return params
 


pdoc/test/__init__.py
```diff
@@ -753,6 +753,12 @@ def test_Function_params(self):
                              lambda a=os.environ, b=sys.stdout: None)
         self.assertEqual(func.params(), ['a=os.environ', 'b=sys.stdout'])
 
+        # This is python 3.8 specific
```
kernc, line 756: Make this a separate test below and `@unittest.skipIf(sys.version_info < (3, 8))`. In the test, you can use `eval()`.

davfsa: Will do

pdoc/test/__init__.py
```diff
@@ -802,6 +802,12 @@ class get_sample(repeat):
         self.assertEqual(pdoc.Function('get_sample', mod, get_sample).return_annotation(),
                          'Tuple[int,\xa0float]')
 
+    @unittest.skipIf(sys.version_info < (3, 8), "This will fail if not using python3.8+")
```
kernc, line 805: Make the warning about the feature, e.g. 'positional-only arguments unsupported in < py3.8'

pdoc/test/__init__.py
```diff
@@ -802,6 +802,12 @@ class get_sample(repeat):
...
+    def test_test_Function_params_python38_specific(self):
+        mod = pdoc.Module(pdoc)  # noqa: F841
+        func = eval("pdoc.Function('f', mod, lambda a, /, b: None)")
```
kernc, line 808: Avoid noqa by only eval-ing the critical section (lambda).

pdoc/test/__init__.py
```diff
@@ -802,6 +802,12 @@ class get_sample(repeat):
...
+        self.assertEqual(func.params(), ['a', '/', 'b'])
```
kernc, line 809: Add another test with `lambda a, /: None`.

pdoc/test/__init__.py
davfsa: Changed

pdoc/test/__init__.py
davfsa: Changed

pdoc/test/__init__.py
davfsa: Added :)

pdoc/test/__init__.py
```diff
@@ -802,6 +802,15 @@ class get_sample(repeat):
...
                          'Tuple[int,\xa0float]')
 
+    @unittest.skipIf(sys.version_info < (3, 8), "positional-only arguments unsupported in < py3.8")
+    def test_test_Function_params_python38_specific(self):
+        mod = pdoc.Module(pdoc)
+        func = pdoc.Function('f', mod, eval("lambda a, /, b: None"))
+        self.assertEqual(func.params(), ['a', '/', 'b'])
+
+        func = pdoc.Function('f', mod, eval("lambda a, /: None"))
+        self.assertEqual(func.params(), ['a'])
```
kernc, line 812: Yeah, looks good, except this is a no go. :smile: It should match `['a', '/']`, to be precise.

davfsa: Fixed

kernc:
Thanks!

PR #198, opened 2020-05-13
reStrucutured text :param: :return: support

### LINKED ISSUE #110: Feature Request: Support reStructuredText-style docstrings
```python
def func(s, test):
    """
    Returns `True` if the string length is greater than the given integer.
    
    :param s: str
    :param test: int
    :return: bool
    """
    
    pass
```

### Expected Behavior

Something like that:
image

### Actual Behavior
image

### Additional info

- pdoc version: 0.6.3

SmirnGreg:
Dear @kernc, maintainers,

For issue #110 I have implemented quite a dirty hack that adds the reSt `:param x:` and `:return:` directives replacing them with the Google-style docstring.

This does not break any of the tests, but I have not implemented a specific test for it, while it works in my case.

Fixes #110

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -129,6 +129,7 @@ class _ToMarkdown:
     — they are all perfect!) an insta-preview tool such as RegEx101.com
     will come in handy.
     """
+
     @staticmethod
     def _deflist(name, type, desc):
         """
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -129,7 +129,6 @@ class _ToMarkdown:
     — they are all perfect!) an insta-preview tool such as RegEx101.com
     will come in handy.
     """
-
     @staticmethod
     def _deflist(name, type, desc):
         """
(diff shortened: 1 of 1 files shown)


kernc:
Doesn't seem cumbersome a transliteration, given its brevity so far. However, I'd much prefer it with the following points addressed:
* [ ] the whole transliteration a _separate_ procedure in e.g. `_ToMarkdown.sphinx_params()`,
* [ ] the procedure returning valid, complete _markdown_, not Google `Args:` (permitted to call through `_ToMarkdown.google()` if needed),
* [ ] the implementation based on `re.sub()` instead of `str.replace()`,
* [ ] accounting for [other Sphinx keywords](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) (`param`, `type`, `raises`, `return`, `rtype`),
* [ ] added a unit test.

kernc (changes requested):
I don't blame you if the above seems like too much work. It'd be a nice addition, though.

PR #202, opened 2020-05-20
Use non-breaking hyphen in ‑> arrow (follow up #51)

laowantong:
The ASCII sequence `->` which has replaced the Unicode arrow character in #51 may break when it's near the last column:

image

Unicode provides a non-breaking hyphen (U+2011) which should look the same, but without this disadvantage: `‑>`.

### DIFF
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -107,7 +107,7 @@
     <dt id="${f.refname}"><code class="name flex">
         <%
             params = ', '.join(f.params(annotate=show_type_annotations, link=link))
-            return_type = get_annotation(f.return_annotation, ''\N{non-breaking hyphen}>'')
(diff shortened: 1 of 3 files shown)


pdoc/templates/html.mako
```diff
@@ -107,7 +107,7 @@
     <dt id="${f.refname}"><code class="name flex">
         <%
             params = ', '.join(f.params(annotate=show_type_annotations, link=link))
-            return_type = get_annotation(f.return_annotation, '->')
+            return_type = get_annotation(f.return_annotation, '‑>')
```
kernc, line 110: Sorry for the late review. Could we make these explicit into `'\N{non-breaking hyphen}>'`?

laowantong: Ah, good idea! I wasn't aware of this escape sequence.

kernc:
Thanks!

Issue #96, opened 2019-08-22
Avoid URL manipulation inside code blocks


AlexFiliakov:
### Expected Behavior
When a docstring has URL inside a code block, display the text as-is without adding any extraneous characters.

### Actual Behavior
Inside code blocks, the URL gets surrounded with <>

(If the code block containing the URL is intended to be copy-pasted somewhere, the appended <> causes problems)

### Steps to Reproduce

1. set your docstring to """For questions ```refer to https://www.google.com```"""
2. generate pdoc3
3. Documentation will say: For questions ```refer to <https://www.google.com```>

(I don't want <> at all, though it also appends > outside of the code block)

### Additional info

- pdoc version: 0.6.3

kernc:
Agreed, code blocks should remain verbatim. Unfortunately, pdoc uses regexes to preprocess raw text into markdown, so detecting and excepting for code blocks might be tricky ...
https://github.com/pdoc3/pdoc/blob/9d71e3269f32c00828d7698d07ca2df0ef8007cc/pdoc/html_helpers.py#L318-L320
PR welcome!

PR #203, opened 2020-05-24
Display link text correctly

### LINKED ISSUE #201: Link text displayed incorrectly
The problem with displaying a link appears in the markdown file. To be more precise, once the link text is placed into [] and https or http is included, this link text is followed by the left angle bracket in generated documentation and leads to a non-existing page. In addition, the right angle bracket is added at the end of the reference.

### Expected Behavior

The link text which starts with https or http is correctly displayed in documentation - exactly the same as provided in [], i.e. :
- link text placed in [] in markdown file: `[https://google.com](https://google.com)`
- link text displayed in the documentation: `https://google.com`

### Actual Behavior

The text link displayed in documentation contains extra left angle bracket:

`<https://google.com`

What is more, the reference ends with the right angle bracket and as a result, it redirects to the page which does not exist:

`https://google.com/>`

### Steps to Reproduce

1. Create markdown file i.e. `readme.md` which contains:
```
[https://google.com](https://google.com)
```
2. Create python file i.e. `example.py` which contains:
```
"""
.. include:: ./readme.md
"""
```
3. Run `$ pdoc3 --http localhost:8080 example`
4. Open `$ http://localhost:8080`

image

image

### Additional info

1. Operating system: Ubuntu 19.10
2. pdoc version: [pdoc 0.8.2.dev4+g048919e](https://github.com/pdoc3/pdoc/commit/048919ece483e0609094cee2460d3467d2e0efc3)
3. Python version: 3.8.0

ButterflyBug:
In generated documentation, the link text placed into [] is not followed
by left angle bracket and it redirects to the given website.

Closes #201

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -361,9 +361,28 @@ class _ToMarkdown:
     @staticmethod
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
+        pattern = re.compile(r"""
+            ((?P<code_span>``.*``|`.*`)      # finds backticks
+            |
+            (?P<markdown_link>\[.*\]\(.*\))  # gets inline link
+            |
+            (?<![<\"\'])                     # does not start with <, ", '
+            (?P<url>(?:http|ftp)s?:\/\/      # url with protocol
+            [^>\s()]+                        # url part before any (, )
+            (?:\([^>\s)]*\))*                # optionally url part within parentheses
+            [^>\s)]*                         # url part after any )
+            ))""", re.VERBOSE)
+
+        def replace(match):
+            groups = match.groupdict("")
+
+            if groups["url"]:
+                return "<" + groups["url"] + ">"
+            else:
+                return match.group()
+
         with _fenced_code_blocks_hidden(text) as result:
-            result[0] = re.sub(r'(?<![<"\'])(\s*)((?:http|ftp)s?://[^>)\s]+)(\s*)',
-                               r'\1<\2>\3', result[0])
+            result[0] = re.sub(pattern, replace, result[0])
         text = result[0]
         return text
 
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -362,7 +362,10 @@ class _ToMarkdown:
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
         pattern = re.compile(r"""
-            ((?P<code_span>``.*``|`.*`)      # finds backticks
+            (?P<code_span>                   # matches whole code span
+                (?<!`)(?P<fence>`+)(?!`)     # a string of backticks
+                .*?
+                (?<!`)(?P=fence)(?!`))
             |
             (?P<markdown_link>\[.*\]\(.*\))  # gets inline link
             |
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -367,25 +367,19 @@ class _ToMarkdown:
                 .*?
                 (?<!`)(?P=fence)(?!`))
             |
-            (?P<markdown_link>\[.*\]\(.*\))  # gets inline link
+            (?P<markdown_link>\[.*?\]\(.*\))  # matches whole inline link
             |
             (?<![<\"\'])                     # does not start with <, ", '
-            (?P<url>(?:http|ftp)s?:\/\/      # url with protocol
-            [^>\s()]+                        # url part before any (, )
-            (?:\([^>\s)]*\))*                # optionally url part within parentheses
-            [^>\s)]*                         # url part after any )
-            ))""", re.VERBOSE)
-
-        def replace(match):
-            groups = match.groupdict("")
-
-            if groups["url"]:
-                return "<" + groups["url"] + ">"
-            else:
-                return match.group()
+            (?P<url>(?:http|ftp)s?://        # url with protocol
+                [^>\s()]+                    # url part before any (, )
+                (?:\([^>\s)]*\))*            # optionally url part within parentheses
+                [^>\s)]*                     # url part after any )
+            )""", re.VERBOSE)
 
         with _fenced_code_blocks_hidden(text) as result:
-            result[0] = re.sub(pattern, replace, result[0])
+            result[0] = pattern.sub(
+                lambda m: ('<' + m.group('url') + '>') if m.group('url') else m.group(),
+                result[0])
         text = result[0]
         return text
 
--- a/pdoc/test/__init__.py
+++ b/pdoc/test/__init__.py
@@ -1339,29 +1339,48 @@ x =</p>'''
 
     def test_urls(self):
         text = """Beautiful Soup
-http://www.foo.bar
-http://www.foo.bar?q="foo"
 <a href="https://travis-ci.org/cs01/pygdbmi"><img src="https://foo" /></a>
 <https://foo.bar>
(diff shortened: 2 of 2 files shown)


pdoc/html_helpers.py
```diff
@@ -361,9 +361,16 @@ def doctests(text):
     @staticmethod
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
+        pattern = (
+            r'(?<![<"[(\'])'       # does not start with <, ", \[, \(, '
+            r'(\s*)'               # can contain whitespace
+            r'((?:http|ftp)s?://'  # protocol name
+            r'[^>)\s]+)'           # >, \), whitespace indicate end of the url
```
kernc, line 368: `\)` won't work for an URL such as `https://en.wikipedia.org/wiki/Orange_(software)`?

pdoc/html_helpers.py
```diff
@@ -361,9 +361,16 @@ def doctests(text):
     @staticmethod
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
+        pattern = (
+            r'(?<![<"[(\'])'       # does not start with <, ", \[, \(, '
```
kernc, line 365: `\(` won't work for URLs in unaesthetic yet plausible parentheses: `... find more info elsewhere (https://url).`?

pdoc/test/__init__.py
```diff
@@ -1365,6 +1365,12 @@ def test_urls(self):
         html = to_html(text)
         self.assertEqual(html, expected)
 
+    def test_url_in_link_text(self):
+        text = "[https://google.com](https://google.com)"
```
kernc, line 1369: Why not add this example into `test_urls()` above, which already tests several cases. Maybe also cover border cases from the other two comments? :sweat_smile:

ButterflyBug:
Hi, I've just updated my PR :smiley: Let me know if you make up some another suspicious type of URLs :smile:

pdoc/html_helpers.py
```diff
@@ -361,9 +361,17 @@ def doctests(text):
     @staticmethod
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
+        pattern = (
+            r'(?<![<"[\](\'])'                      # does not start with [, <, ", ], (, '
+            r'('
+            r'((?P<surround>\()|(\(?))'             # link can be surrounded with (
+            r'(?P<url>(?:http|ftp)s?://[^>\s]+)'    # url with proocol
+            r'(?(surround)(\))|\)?)'                # if url starts with ( it ends with )
```
kernc, line 369: Love the commented formatting! :heart_eyes: 

After sleeping on it, I think we can complicate it further. :sweat_smile: Maybe the regex should except only on characters `<"'` as before, and when within markdown link definition, either text part (i.e. somewhere after a non-closed `[`) or URL part (after `](`). That seems like a catch-_less_ approach? 

I'm also certain we should avoid links in inline code, i.e. [non-closed series of `` ` ``](https://daringfireball.net/projects/markdown/syntax#code) backticks. Refs: https://github.com/pdoc3/pdoc/issues/96

pdoc/html_helpers.py
```diff
@@ -361,9 +361,29 @@ def doctests(text):
...
+        pattern = (
+            r'(?P<backtick>`.*`)'                   # finds backticks
+            r'|'
+            r'(?P<markdown_link>(\[.*\])(\(.*\)))'  # gets inline link
+            r'|'
+            r'(?<![<"\'])'                          # does not start with <, ", '
+            r'('
+            r'((?P<left>\()|(\(?))'                 # link can be surrounded with (
+            r'(?P<url>(?:http|ftp)s?://[^>\s]+)'    # url with protocol
+            r'(?(left)(?P<right>\))|\)?)'           # if url starts with (, it ends with )
```
kernc, line 373: Sorry, I have reservations about this assumption. Besides failing to match
> something something (http://foo and http://bar)

what's it do?

kernc: A little different regex seems to pass including this far-fetched test: https://regex101.com/r/CJgYsF/3.
What do you think?

I wonder what regex GitHub/Facebook uses. They seem to work well. :thinking: 
> http://url)
> http://url()
> http://url.
> http://url.url
> http://url,
> http://url,asd

ButterflyBug: Test added, your version of regex - checked. All look like working correctly :smile_cat: Thanks a lot for your idea and..another challenge :smile:

pdoc/html_helpers.py
```diff
@@ -361,9 +361,29 @@ def doctests(text):
     @staticmethod
     def raw_urls(text):
         """Wrap URLs in Python-Markdown-compatible <angle brackets>."""
+        pattern = (
+            r'(?P<backtick>`.*`)'               # finds backticks
```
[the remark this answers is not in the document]
kernc: Also rename the matching group to `code_span`. Can you also make sure it matches [inline code spans with longer fences](https://spec.commonmark.org/0.29/#code-spans)? This test, basically:
```markdown
lala ``x ` http://foo`` http://bar `http://foo`
```

pdoc/html_helpers.py
```diff
@@ -361,9 +361,29 @@ def doctests(text):
...
+            r')'
+        )
+
+        def replace(match):
+            groups = match.groupdict("")
+
+            if groups["backtick"] or groups["markdown_link"]:
+                return match.group()
+            else:
+                return "<" + groups["url"] + ">"
```
kernc, line 383: Also maybe invert this `if` to check for `url` group first? I don't like `else` on short-circuit returns. This now almost fits on one line!

pdoc/test/__init__.py
```diff
@@ -1352,7 +1352,16 @@ def test_urls(self):
...
 http://url.com
-```"""
+```
+[https://google.com](https://google.com)
+[https://en.wikipedia.org/wiki/Orange_(software)](https://en.wikipedia.org/wiki/Orange_(software))
+https://en.wikipedia.org/wiki/Orange_(software)
+(https://google.com)
+[Check https://google.com here](https://google.com)
+`https://google.com`
+(http://foo and http://bar)
```
kernc, line 1362: Finally, maybe split the tests by matching and non-matching together. Yea, I'm pushing it. :sweat_smile:

pdoc/html_helpers.py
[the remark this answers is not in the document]
ButterflyBug: haha no worries ;D I learn a lot from these whims. Besides, thanks to them our regex looks better and better. :smile: I agree that using `re.VERBOSE`   would be a prettier solution, thanks for that comment :smile:

pdoc/test/__init__.py
ButterflyBug: I'm not sure if I understood that correctly. :thinking:  You mean to split `(http://foo and http://bar)` into two separated tests like `(http://foo something)` and `(something http//:bar)?` :smile:

kernc: I meant to split all these test lines into those that match (and should produce `<url>`) and those that don't (shouldn't).

kernc:
I pushed some pedantic touch-ups I didn't feel like voicing aloud. This now looks good and should hopefully stick as far as wrapping raw URLs goes. :crossed_fingers: Much appreciated!

If you care to make other regexes in the module more reader-friendly, or if you find other bugs, obviously welcome. :grin: Thanks!

PR #209, opened 2020-06-08
Offline search with lunr.js

### LINKED ISSUE #184: ENH: Offline search using JS index
At the time of opening this issue, the search bar consists of a google search bar.

It would be nice for pdoc to have a offline search bar (something like what sphinx has) to have a faster, non third party,  offline way of searching the docs.

davfsa:
This will replace the current google search bar with an offline search bar using lunr.js

Closes #184


davfsa:
@kernc Ready for a review :)

davfsa:
Fixed merge conflicts

pdoc/cli.py
```diff
@@ -337,7 +344,9 @@ def recursive_write_files(m: pdoc.Module, ext: str, **kwargs):
         raise
 
     for submodule in m.submodules():
-        recursive_write_files(submodule, ext=ext, **kwargs)
+        recursive_write_files(submodule, ext=ext, _rendered_modules=_rendered_modules, **kwargs)
+
+    return _rendered_modules
```
kernc, line 349: Instead of passing `_rendered_modules` around, you can just return a list and then merge with submodules' lists.
```py
    submodule_files = recursive_write_files(...)
    rendered_modules.extend(submodule_files)

return rendered_modules
```
Also let's maybe call the variable `written_files`.

pdoc/cli.py
```diff
@@ -498,7 +507,50 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
     for module in modules:
         if args.html:
             _quit_if_exists(module, ext='.html')
-            recursive_write_files(module, ext='.html', **template_config)
+            rendered_modules = recursive_write_files(module, ext='.html', **template_config)
```
kernc, line 510: definitely move this hunk into a separate function

pdoc/cli.py
```diff
@@ -498,7 +507,50 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
...
+
+            if path.dirname(module_path(module, ext='')) == args.output_dir:
+                continue
```
kernc, line 513: explanation comment here

pdoc/templates/config.mako
```diff
@@ -42,16 +42,18 @@
     # tracking id (UA-XXXXXX-Y).
     google_analytics = ''
 
-    # If set, insert Google Custom Search search bar widget above the sidebar index.
-    # The whitespace-separated tokens represent arbitrary extra queries (at least one
-    # must match) passed to regular Google search. Example:
-    #search_query = 'inurl:github.com/USER/PROJECT  site:PROJECT.github.io  site:PROJECT.website'
-    search_query = ''
```
kernc, line 49: No need to remove this option. It's a viable alternative.

kernc, line 49: No need to remove this option. It's a viable alternative.

pdoc/templates/config.mako
```diff
@@ -42,16 +42,18 @@
...
     # or you need to use raw r-strings.
     latex_math = False
+
+    # The fuzziness of the web search. Setting it to 0 will allow you to specify
+    # the fuzziness in the query.
+    #
+    # WARNING: SETTING THIS NUMBER TOO HIGH WILL GIVE TOO MANY RESULTS AND MAKE
+    # YOUR BROWSER REALLY SLOW OR EVEN FREEZE IT. WE RECOMEND YOU KEEP THIS 
+    # BETWEEN 0 AND 3.
+    fuzziness = 1
```
kernc, line 58: So `lunr_search_fuzziness`?

And point to futher docs, `https://lunrjs.com/guides/searching.html#fuzzy-matches`.

pdoc/templates/html.mako
```diff
@@ -288,72 +288,74 @@
...
-    % if submodules:
-    <li><h3><a href="#header-submodules">Sub-modules</a></h3>
-      <ul>
-      % for m in submodules:
-        <li><code>${link(m)}</code></li>
-      % endfor
-      </ul>
-    </li>
-    % endif
+    <form action="${'../'*module.name.count('.')}search.html" method="GET">
```
kernc, line 291: Is it possible to load search results in an overlay iframe, without leaving the active page, like the current search behaves?

I think it's acceptable for this to only work with JavaScript, degrading gracefully to loading _search.html_ page when JS is unavailable.

pdoc/templates/html.mako
```diff
@@ -288,72 +288,74 @@
...
+              if sort_identifiers:
+                members = sorted(members)
+          %>
+          % if members:
+            ${show_column_list(members)}
+          % endif
+          </li>
+        % endfor
+        </ul>
+      </li>
```
kernc, line 355: Not sure what changed here, but if it's just improved indentation, no need to do that in this exact one big ugly commit.

pdoc/templates/html.mako
```diff
@@ -416,15 +412,24 @@
...
+      </article>
+      ${module_index(module)}
+    % endif
+
   % else:
-    <article id="content">
-      ${show_module(module)}
-    </article>
-    ${module_index(module)}
+    <%namespace name="search" file="search.mako"/>
```
kernc, line 428: And going the way of overlay, _search.mako_ then need not inherit from _html.mako_, particularly since it only refers to `${fuzziness}` from _config.mako_.

pdoc/templates/search.mako
```diff
@@ -0,0 +1,99 @@
...
+<article id='content'>
+    <h1>Search the docs</h1>
+    <h3 id="search-status"></h3>
+    <div id="search-results"></div>
+</article>
+</%def>
+
+## Version of lunr.js is frozen to 2.3.8, update here if necesary
+<%def name="js()">
+<script src="https://unpkg.com/lunr@2.3.8/lunr.js"></script>
```
kernc, line 11: We subscribe to [CDNjs resources](https://cdnjs.com/libraries/lunr.js) elsewhere, along with SRIs. See existing `<script src>` tags.

pdoc/test/__init__.py
```diff
@@ -97,8 +97,10 @@ class CliTest(unittest.TestCase):
     ALL_FILES = [
         'example_pkg',
         'example_pkg/index.html',
+        'example_pkg/index.js',
         'example_pkg/index.m.html',
         'example_pkg/module.html',
+        'example_pkg/search.html',
```
kernc, line 103: Once lunr search is made optional, you can just create a new separate and minimal test for it.

These regression CLI tests were used mainly at initial overhaul of pdoc code.

pdoc/cli.py
```diff
@@ -498,7 +507,50 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
...
+            for rendered_module in rendered_modules:
+                info = {}
+                info['ref'] = rendered_module.url()[len(module.name)+1:]
+                info['name'] = rendered_module.name
+                info['refname'] = rendered_module.refname
+                info['docstring'] = rendered_module.docstring
+
+                index.append(info)
+
+                for mod in (
```
kernc, line 526: `dobj` for "documentation or Doc object" is used elsewhere.

pdoc/cli.py
```diff
@@ -498,7 +507,50 @@ def docfilter(obj, _filters=args.filter.strip().split(',')):
...
+                    + rendered_module.functions()
+                ):
+                    info = {}
+                    info['ref'] = mod.url()[len(module.name)+1:]
+                    info['name'] = mod.name
+                    ref = mod.refname + ('()' if isinstance(mod, pdoc.Function) else '')
+                    info['refname'] = ref
+                    info['docstring'] = mod.docstring
+
+                    index.append(info)
```
kernc, line 538: So this just dumps every raw docstring of every documented object into this file. Couldn't we opt instead for creating a [pre-built index](https://lunrjs.com/guides/index_prebuilding.html)?

pdoc/templates/html.mako
davfsa: `search.html` is basically also JS, so if JS is unavailable, the search will not work. I thoughtI I added a warning for missing JS but looks like I didnt (going to do so).

Still want to go with the iframe? I didnt go with this since search imports the `index.js` which might be really big for some projects, so I would rather keep it all in `search.html` so it only gets loaded when needed.

pdoc/templates/search.mako
davfsa: So i should use this? `<script src="https://cdnjs.cloudflare.com/ajax/libs/lunr.js/2.3.8/lunr.min.js></script>`

pdoc/cli.py
davfsa: Do you want me to change `mod` by `dobj`

pdoc/templates/html.mako
kernc: Fairly easy to load additional heavy resources only when needed: https://stackoverflow.com/a/14521482/1090455

pdoc/cli.py
kernc: From the naming point of view, it makes some sense. The objects iterated over are variables, classes and functions, specifically not modules.

pdoc/cli.py
davfsa: Removed this in the end since it was hiding a bug inside. I deal with this another way now

pdoc/cli.py
davfsa: Found a [lunr python module](https://pypi.org/project/lunr/) which I will be using to generate the pre-built index

pdoc/templates/html.mako
davfsa: This was a really ugly `if` statement, ended up just doing it another easier way :)

pdoc/cli.py
davfsa: Looks like im not going to be able to implement the pre-built index because as I create the index I also store all the info I pass to `documentsIndex` to be able to reference it later:
```js
        var idx = lunr(function () {
                this.ref('ref');
                this.field('refname', {'boost':10});
                this.field('name', {'boost': 5});
                this.field('docstring');
                this.metadataWhitelist = ['position'];

                index.forEach(function(doc) { 
                    this.add(doc);
                    documentsIndex[doc.ref] = {
                        'refname': doc.refname,
                        'name': doc.name,
                        'docstring': doc.docstring,
                    };
                }, this);
            });
```
And also, since when pre-building you are not able to specify the metadata i want to receive from the queries, I dont think this is possible.

pdoc/templates/html.mako
davfsa: I rather keep it as it a difference file because i think its a whole different module. It makes it really easy to load the search if they replace the `html.mako` or replace the search it if they need it.

pdoc/templates/html.mako
davfsa: I really dont see the point for this. I could implement it, but its a big useless, since you dont get anything out of it. Something I could do that I have been thinking about is that when you go into `search.html` and you search from there, it doesnt refresh the page, but just show you the new results

pdoc/templates/html.mako
kernc: Absolutely, that's what I meant: keep two separate templates. That's why if you load the results in an iframe [as discussed above](https://github.com/pdoc3/pdoc/pull/209#discussion_r447332888), you don't even need to hack into _html.mako_ to reuse its layout and styles, because you can just make a simple results page, with layout and styles pertaining to its results only, i.e.:
Screenshot_2020-07-02 pdoc API documentation

pdoc/cli.py
```diff
@@ -76,6 +77,14 @@
     type=str,
     help=argparse.SUPPRESS,
 )
+aa(
+    "--html-search-type",
+    type=str,
+    metavar='STRING',
+    default=None,
+    help="When set, type of search to use. One of 'google' or 'lunr'. "
+         "If unset, no search will be used.",
```
kernc, line 86: No, currently various pdoc's toggles are configured through _config.mako_. That may not be the best approach, but it's certainly preferable to a deluge of CLI flags.

Let's puts `lunr_search_config` or some such object inside _config.mako_, and then if set (see how other existing config flags are preset), enable lunr search. If both Google and lunr search are enabled, we include/inject code for both — the user knows what they are doing.

The neat thing with config tunables is that they can be easily set with `--config` CLI flag. Please investigate.

pdoc/cli.py
```diff
@@ -362,6 +377,50 @@ def _warn_deprecated(option, alternative='', use_config_mako=False):
     warn(msg, DeprecationWarning, stacklevel=2)
 
 
+def _generate_lunr_search(top_module, written_files, template_config) -> None:
+    # Generate index.js for search
+    index = []
+    for written_file in written_files:
```
kernc, line 383: So these are modules, not file strings. Don't take my suggestions so literally. :sweat_smile: 
I think
```py
for module in modules:
```
will be just fine here now that it's a separate function that encompasses what those modules are. Prefer it brief and exact. :+1:

pdoc/cli.py
```diff
@@ -362,6 +377,50 @@ def _warn_deprecated(option, alternative='', use_config_mako=False):
...
+        index.append(info)
+
+        for dobj in (
+            written_file.variables()
+            + written_file.classes()
+            + written_file.functions()
+        ):
+            info = {}
+            info['ref'] = (dobj.url()[len(top_module.name)+1:]
+                           if top_module.is_package else dobj.url())
```
kernc, line 400: What does this do?

pdoc/cli.py
```diff
@@ -362,6 +377,50 @@ def _warn_deprecated(option, alternative='', use_config_mako=False):
...
+            info['name'] = dobj.name
+            ref = dobj.refname + ('()' if isinstance(dobj, pdoc.Function) else '')
+            info['refname'] = ref
+            info['docstring'] = dobj.docstring
+
+            index.append(info)
+
+    main_path = path.join(args.output_dir,
+                          *top_module.name.split('.') if top_module.is_package else '')
```
kernc, line 409: And what does this do?

pdoc/templates/config.mako
kernc: The rename is ok, but let's handle the deprecation in `_render_template()`:
https://github.com/pdoc3/pdoc/blob/04960e41b9bb598b664a50078f4c233817326170/pdoc/__init__.py#L112-L123
`search_query` should still work, but issue a warning.

pdoc/templates/html.mako
kernc: Two separate templates, but not tangled into one another. _html.mako_ is heavy enough as it is ...

davfsa: Ah, ok, got you, will implement it :+1:

pdoc/cli.py
davfsa: The reason behind having it like this is because I need to know this information in this file since we never load `config.mako` here. Now that I think about it, I can just read it. Will add it

pdoc/cli.py
davfsa: (Will add this into a comment) `search.html` shoud sit at `/output-dir/package/` but since when you build the docs for one file (lets call it `module.py`) you dont get that subfolder, `path.join(args.output_dir, *top_module.name.split('.')` alone would return `/output-dir/module`. Which is not what we want. So this basically checks if its a package, if so, go to the subfolder, if not, stay in the upper folder

pdoc/cli.py
davfsa: (Will add this into a comment) The references should be made from `search.html`, meaning that they have to be in the subfolder (if any). This means that we will need to remove the subfolder. This basically checks if its a package, remove the subfolder name + 1 (`/`) from the url, if not, leave it alone.

*Note: while writing this i realiced that this wont actually work on windows, so that is something that I will have a look at*

pdoc/templates/html.mako
davfsa: So, after having a lot of thoughts and experiments on this, I havent been able to implement it using an iframe due to lots of bugs. I tried to implement it how google does it, but they use full url paths and there is no way we can do that, causing a bug that I cant fix.

Because of this I have decided to keep it how its now. I have changed the code the the .mako files to have search.mako and html.mako separate (not how it was before). Hope its ok :smile:

pdoc/templates/html.mako
davfsa: https://github.com/pdoc3/pdoc/pull/209#discussion_r449752002

pdoc/cli.py
kernc: Metadata we want is the match position? Then indeed not possible (https://github.com/yeraydiazdiaz/lunr.py/issues/77). :disappointed:

pdoc/templates/search.mako
```diff
@@ -0,0 +1,153 @@
...
+        results.forEach(function(result) {
+            var liLink = '<a href="' + result.ref + '">' + documentsIndex[result.ref].refname + '</a>';
+
+            Object.keys(result.matchData.metadata).forEach(function (term) {
+                if (!('name' in result.matchData.metadata[term] || 'refname' in result.matchData.metadata[term])) {
+                    var docstring = documentsIndex[result.ref].docstring;
+
+                    result.matchData.metadata[term].docstring.position.forEach(function(positions) {
+                        var start;
+                        if (positions[0] - 180 > 0) {
```
kernc, line 118: What is `180`? If this is some arbitrary text context padding in character units, make a variable for it:
```js
const TEXT_CONTEXT_CHARS = 180;
```
Better avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

pdoc/templates/search.mako
```diff
@@ -0,0 +1,153 @@
...
+</footer>
+
+<script src="index.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/lunr.js/2.3.8/lunr.js" integrity="sha256-JZUIb2pF9vF82f9JemSl6XISUBX4tvjrprJM21J63G8=" crossorigin></script>
+<script>
+var searchStatus = document.getElementById('search-status');
+var searchResults = document.getElementById('search-results');
+var searchInput = document.getElementById('search-box');
+var idx = null;
+var documentsIndex = null;
```
kernc, line 57: IIUC, `documentsIndex` is nearly identical to `index` from _index.js_, except one is a mapping and the other a plain list. You could avoid the list and maintain just a mapping _refname_→_doc_, setting lunr `ref` to the same _refname_, no?

pdoc/cli.py
```diff
@@ -362,6 +369,56 @@ def _warn_deprecated(option, alternative='', use_config_mako=False):
...
+            module.variables()
+            + module.classes()
+            + module.functions()
+        ):
+            info = {}
+            url = dobj.url()
+            if top_module.is_package:  # Reference from subfolder if its a package
+                url = "/".join(url.split("/")[1:])
+
+            info['ref'] = url
```
kernc, line 398: Why, really, not just use `Doc.refname` as the object's `id`?

pdoc/templates/html.mako
```diff
@@ -412,6 +414,15 @@
...
   % endif
 
+  % if google_search_query:
+    <link rel="preconnect" href="https://www.google.com">
+    <script async src="https://cse.google.com/cse.js?cx=017837193012385208679:pey8ky8gdqw"></script>
+    <style>
+        .gsc-control-cse {padding:0 !important;margin-top:1em}
+        body.gsc-overflow-hidden #sidebar {overflow: visible;}
+    </style>
+  % endif
```
kernc, line 424: Maybe revert moving this hunk. Only related changes are appropriate for a _clean_ PR.

pdoc/templates/html.mako
```diff
@@ -276,7 +275,7 @@
   </section>
 </%def>
 
-<%def name="module_index(module)">
+<%def name="module_index(module, search_tab=False)">
```
kernc, line 278: So what's with this `search_tab`?

pdoc/cli.py
```diff
@@ -362,6 +369,56 @@ def _warn_deprecated(option, alternative='', use_config_mako=False):
...
+    for module in modules:
+        info = {}
+        url = module.url()
+        if top_module.is_package:  # Reference from subfolder if its a package
+            url = "/".join(url.split("/")[1:])
+
+        info['ref'] = url
+        info['name'] = module.name
+        info['refname'] = module.refname
+        info['docstring'] = module.docstring
```
kernc, line 384: If we can't prebuild, wouldn't it make sense to at least strip the plain text docstrings of extraneous [stop words](https://en.wikipedia.org/wiki/Stop_words), like [these](https://gist.github.com/sebleier/554280)? Skipping all likely irrelevant words should certainly bring down the size of the index considerably ...

pdoc/templates/html.mako
davfsa: Forgot to remove it

pdoc/templates/config.mako
```diff
@@ -42,16 +42,28 @@
...
+
+    # The fuzziness of the lunr search. Setting it to 0 will allow you to specify
+    # the fuzziness in the query by using '~'.
+    #
+    # https://lunrjs.com/guides/searching.html#fuzzy-matches
+    #
+    # WARNING: SETTING THIS NUMBER TOO HIGH WILL GIVE TOO MANY RESULTS AND MAKE
+    # YOUR BROWSER REALLY SLOW OR EVEN FREEZE IT. WE RECOMEND YOU KEEP THIS 
+    # BETWEEN 0 AND 2.
+    lunr_search_fuzziness = 1
```
kernc, line 68: Can we also make this just a single config parameter. I suspect we might add other toggles to it in the future because it's such a large subsystem. The setting can be a dict:
```py
# Enables lunr.js search ...
#lunr_search = {
#   'fuzziness': 2,
#   'stopwords': [ ... ],
#}
lunr_search = None
```

pdoc/templates/config.mako
```diff
@@ -42,16 +42,28 @@
...
-    search_query = ''
-
     # If set, render LaTeX math syntax within \(...\) (inline equations),
     # or within \[...\] or $$...$$ or `.. math::` (block equations)
     # as nicely-formatted math formulas using MathJax.
     # Note: in Python docstrings, either all backslashes need to be escaped (\\)
     # or you need to use raw r-strings.
     latex_math = False
+
+    # The whitespace-separated tokens represent arbitrary extra queries (at least one
```
kernc, line 52: This migrated description lost the leading original line:
```
If set, insert Google Custom Search search bar widget above the sidebar index.
```

pdoc/templates/config.mako
davfsa: Ok, will do. 

Also, question on the `stopwords`. Now with this, should I add here in the config?

pdoc/templates/config.mako
davfsa: My bad :sweat_smile:

pdoc/templates/config.mako
kernc: I think an English list is a sensible default, but should probably be configurable.
Indeed, maybe better to define the list elsewhere (_html_helpers_?) and then just import and reference it here ... :thinking:

davfsa: :thinking: This will be challenging to implement. Will have a look

pdoc/cli.py
davfsa: Because the objects `id` doesnt get searched in. I want to be able to do `example_pkg.numpy` and be able to get a result back.

pdoc/templates/search.mako
davfsa: Found a way to do this :+1:

pdoc/cli.py
davfsa: After having a look at this, I think its a bad idea. Several reasons:
1. If we remove stopwords, the search results will not make any sense, since they will be missing the stopwords.
2. Having a look back at the pre-built indexes, they are actually bigger than just storing everything raw as i do now. 
*These are for documentation built from the `example_pkg` in the project*
35.9 kB - Raw index
180.4 kB - Pre-built index

In conclusion, I think the way its set up right now is the best :)

pdoc/cli.py
kernc: But isn't this exactly what fields do?
https://github.com/pdoc3/pdoc/blob/f43772a6c89b8b214bde7a52cfc876164881b99b/pdoc/templates/search.mako#L74-L76

davfsa: Ok, I think i misunderstood you, what do you mean by `objects id`?

davfsa:
@kernc anything else you would like to see changed?

petrisorionel:
when this commit will be merged ?

davfsa:
@kernc Any ETA for when 0.9.0 will be released? Its been really long since the last release.

kernc:
While this PR introduces many great ideas, it was left riddled with places for improvement, and I don't like to merge and release half-baked solutions that have potential to trigger an avalanche of new issues on the bug tracker. :sweat_smile: Admittedly, this should have been ironed out in the review process, but being preoccupied with other interesting stuff, I let this sit for too long and then didn't feel like voicing my bikeshedding dispositions aloud. Apologies.

@davfsa I applied some changes on my checkout that I'd be happy to push on here for you to review, if you permit me so. I think there's a 'Allow edits from maintainers' checkbox somewhere in the right sidebar.

davfsa:
Sure, fine by me :smile: I made sure it's enabled for you to make commits. If you need anything else make sure to let me know :smile:

kernc:
As you can see, I started small, but then managed to reduce the generated "index" size by ~20% and ended up including implementing including the results on the same top page in a modal iframe. Welcome your thoughts.

pdoc/templates/config.mako
kernc: Lunr.js has [its own stop-words filter](https://github.com/olivernn/lunr.js/blob/f9aeea2c592dc08eec67937bcb5b34903e6c04dc/lib/stop_word_filter.js). :+1:

pdoc/cli.py
kernc: I meant something like this: https://github.com/pdoc3/pdoc/pull/209/commits/1a74fb66c9afc4de2100049d5745c524025cc2d3. Note how the generated index is minimal, with much fewer repetitions (names contained in refnames, URLs enumerated and stitched on the fly, ...).

kernc (approved):

davfsa:
So, several things.
1. I dont really like the usage of iframes, they are just too much hassle for a very little improvement (https://stackoverflow.com/a/23178537). I like the way you implemented it but i just think it will be a PITA to maintain. 
2. The point of the `search.html` is for people to be able to search stuff without having to render the documentation. Which, on a slow device and a huge landing page, its greatly appreciated.
3. If a browser with javascript disabled tries to use the search bar, the search bar will lock itself without giving any feedback back to the user. Since you use javascript to load the iframe, this isnt something you can fix. I would recommend having it the way I had it before.
4. Code looks much nicer, thank you a lot for that :sweat_smile: and I like the way you organized it.
5. Input box doesnt reset after iframe close and on page load.
6. Settings config to `{}` should work but it doesnt. This can be fixed by doing `if lunr_search is None:` instead of `if lunr_search:`. This is probably something you should fix because setting it to `{}` will cause the index and search to generate but not the search tab.

pdoc/templates/search.mako
```diff
@@ -107,9 +107,8 @@
                     .filter(({doc}) => doc !== undefined)
                     .map(({doc: {position}}) => {
                         return position.map(([start, length]) => {
-                            const PAD_CHARS = 30
+                            const PAD_CHARS = 30;
                             const end = start + length;
-                            ## TODO: merge overlapping matches
```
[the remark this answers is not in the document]
davfsa: Ah, ok, will re-add the TODO then

pdoc/templates/search.mako
```diff
@@ -54,8 +54,8 @@
     async function build_index() {
         return lunr(function () {
             this.ref('i');
-            this.field('ref', {boost: 5});
-            this.field('name', {boost: 10});
+            this.field('ref', {boost: 10});
+            this.field('name', {boost: 5});
```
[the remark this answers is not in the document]
davfsa: Oh, true, good point

davfsa:
@kernc Fixed all comments :smile:

kernc:
Please permit me to push to this branch again. I have a few final editorial amendments to propose.

davfsa:
> Please permit me to push to this branch again. I have a few final editorial amendments to propose.

Enabled :+1:

kernc:
There. A few of those are nitpicks, but this is how I would have it. Note the renaming of `lunr_config.index_docstrings`.

kernc:
Let's freeze this here for now. We will undoubtedly see new related issues and needed follow ups. :sweat_smile:

Thanks @davfsa for all the hard work and @nekokatt for reviewing.

al-pa:
Hello @davfsa, @kernc ,
When I tried to use offline search functionality I get such error in pop-up window:
```
Error importing module doc-search:

  File "/home/al-pa/localPython/py380/lib/python3.8/site-packages/pdoc/__init__.py", line 225, in import_module
    module = importlib.import_module(module_path)
  File "/home/al-pa/localPython/py380/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'doc-search'

  File "/home/al-pa/localPython/py380/lib/python3.8/site-packages/pdoc/cli.py", line 234, in do_GET
    out = self.html()
  File "/home/al-pa/localPython/py380/lib/python3.8/site-packages/pdoc/cli.py", line 263, in html
    return pdoc.html(self.import_path_from_req_url,
  File "/home/al-pa/localPython/py380/lib/python3.8/site-packages/pdoc/__init__.py", line 176, in html
    mod = Module(import_module(module_name, reload=reload),
  File "/home/al-pa/localPython/py380/lib/python3.8/site-packages/pdoc/__init__.py", line 227, in import_module
    raise ImportError('Error importing {!r}: {}: {}'
ImportError: Error importing 'doc-search': ModuleNotFoundError: No module named 'doc-search'

```
**pip install doc-search**  tell me that
_ERROR: Could not find a version that satisfies the requirement doc-search
ERROR: No matching distribution found for doc-search_

Could you please help me to understand what I am missing?

P.S. 
```
$ pdoc3 --version
pdoc3 0.9.2

$ python --version
Python 3.8.5

```
P.P.S. running pdoc with such command:
`pdoc3 --html --http localhost:1234 -c lunr_search=True <module_name>`

Thank you!

kernc:
@al-pa Feel free to open a new issue about this!

al-pa:
Thank you, issue created: #295

PR #216, opened 2020-06-12
Normalize `.. include::` path on Windos

lluissalord:
This is fixing issue when you have an `include` on an `__init__.py` file which is trying to find a file on a parent directory in Windows.

With the previous implementation, it was mixing `\\` (from Windows) with `../` from include and also the final path was not the correct one.


kernc:
Yeah, Windows is maybe not perfectly covered. :sweat_smile: Thanks!

PR #225, opened 2020-06-30
Test building PDF with pandoc on Travis

kernc:
The test is conditional on an environment variable because it takes as much time as all the other tests combined.


kernc:
Seems [texlive-xetex on Xenial is too old](https://travis-ci.org/github/pdoc3/pdoc/jobs/703553676#L594). ICBA at the moment. Let's revisit when updating CI env, or try a different Latex engine.

PR #242, opened 2020-08-09
Use __init__ instead of __new__ to document class params

### LINKED ISSUE #191: __init__ arguments are being replaced by *args, **kwargs
The following code
```
import plotly.graph_objects as go

class FigureWidget(go.FigureWidget):
    def __init__(self):
        """Documentation."""
        pass
```

Produces the following documentation

Screen Shot 2020-04-28 at 19 14 43

Although I clearly defined that the `__init__` method should have no arguments, it shows `*args` and `**kwargs` in the signature. Are there any caveats in subclassing third-party classes? Also any ideas how to unveil the "real" signature?

RaduG:
Fixes #191 

Using `inspect.signature` to get the signature of a class will return the signature of `__new__`, if implemented. However, I don't think there are necessarily use cases for this when `__init__` is actually implemented. For example:

```python
class A(Generic[T]):
   """doc"""
   def __init__(self, a, b):
        """bar"""
        pass
```

This will show the correct documentation but the signature will be, of course, `*args, **kwargs`. The small proposed change ensures that the signature of `__init__` is used if it is implemented (i.e. `obj.__init__ is not object.__init__`. 

EDIT:
Added an additional test case for when `__new__` is implemented but `__init__` isn't.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1242,7 +1242,15 @@ class Function(Doc):
     @staticmethod
     def _params(doc_obj, annotate=False, link=None, module=None):
         try:
-            signature = inspect.signature(doc_obj.obj)
+            # We want __init__ to actually be implemented somewhere in the
+            # MRO to still satisfy https://github.com/pdoc3/pdoc/issues/124
+            if (
+                inspect.isclass(doc_obj.obj)
+                and doc_obj.obj.__init__ is not object.__init__
+            ):
+                signature = inspect.signature(doc_obj.obj.__init__)
+            else:
+                signature = inspect.signature(doc_obj.obj)
         except ValueError:
             signature = Function._signature_from_string(doc_obj)
             if not signature:
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1248,7 +1248,10 @@ class Function(Doc):
                 inspect.isclass(doc_obj.obj)
                 and doc_obj.obj.__init__ is not object.__init__
             ):
-                signature = inspect.signature(doc_obj.obj.__init__)
+                # Remove the first argument (self) from __init__ signature
+                init_sig = inspect.signature(doc_obj.obj.__init__)
+                init_params = list(init_sig.parameters.values())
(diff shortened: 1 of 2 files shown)


pdoc/test/__init__.py
```diff
@@ -915,7 +928,7 @@ def __init__(self, x):
                 pass
 
         mod = pdoc.Module(pdoc)
-        self.assertEqual(pdoc.Class('C', mod, C).params(), ['x'])
+        self.assertEqual(pdoc.Class('C', mod, C).params(), ['self', 'x'])
```
kernc, line 931: I think we better retain class constructors documented as they are used to instantiate, just the params.

RaduG: Done! Let me know what you think, I reckon `signature.replace` is decent enough in this case.

pdoc/__init__.py
```diff
@@ -1242,7 +1242,18 @@ def params(self, *, annotate: bool = False, link: Callable[[Doc], str] = None) -
...
+            # We want __init__ to actually be implemented somewhere in the
+            # MRO to still satisfy https://github.com/pdoc3/pdoc/issues/124
+            if (
+                inspect.isclass(doc_obj.obj)
+                and doc_obj.obj.__init__ is not object.__init__
+            ):
+                # Remove the first argument (self) from __init__ signature
+                init_sig = inspect.signature(doc_obj.obj.__init__)
+                init_params = list(init_sig.parameters.values())
+                signature = init_sig.replace(parameters=init_params[1:])
```
kernc, line 1254: Can we assert here `'self'` is in fact the first param we're dropping.

RaduG: I'm not sure what the value is. `__init__` will always be a bound method and the name doesn't need to be `self`, so I think it's 100% safe to assume that the first argument is the current instance.

kernc:
Perfect, thanks!

RaduG:
Do you need anything else from me for this one?

kernc:
No. Was just postponing it for the shortly upcoming feature release.

Thanks again for the PR!

PR #249, opened 2020-08-31
BUG: use last __init__ to document class variables

### LINKED ISSUE #248: __init__ with @overload prevents pdoc from documenting instance variables
### Expected Behavior

If a class `@overload`s its `__init__` method, variable docstrings should be taken from the implementation.

### Actual Behavior

Docstrings are taken from the first `@overload`, not the implementation.

### Steps to Reproduce

Minimal example:

    from typing import Union, Literal, overload

    class Example:
        @overload
        def __init__(self, typ: Literal["str"], x: str):
            pass

        @overload
        def __init__(self, typ: Literal["int"], x: int):
            pass

        def __init__(self, typ: Union[Literal["str"], Literal["int"]], x: Union[str, int]):
            self.x = x
            """
            Value provided in the constructor.
            """

### Additional info

Currently, one can work around this limitation by putting the docstring in the first `@overload`ed version; e.g.:

    class Example:
        @overload
        def __init__(self, typ: Literal["str"], x: str):
            self.x = x
            """
            Value provided in the constructor.
            """

        @overload
        def __init__(self, typ: Literal["int"], x: int):
            pass

        def __init__(self, typ: Union[Literal["str"], Literal["int"]], x: Union[str, int]):
            self.x = x

However, this is less than ideal.

- pdoc version: 0.8.1

nosewings:
Fixes #248.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -262,16 +262,11 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
             tree = tree.body[0]  # ast.parse creates a dummy ast.Module wrapper
 
             # For classes, maybe add instance variables defined in __init__
-            init_nodes = [
-                node
-                for node in tree.body
-                if isinstance(node, ast.FunctionDef) and node.name == "__init__"
-            ]
-            if init_nodes:
-                # Get the *last* init node in case it is preceded by @overrides.
-                instance_vars, _ = _pep224_docstrings(
-                    doc_obj, _init_tree=init_nodes[-1]
-                )
+            # Get the *last* __init__ node in case it is preceded by @overloads.
+            for node in reversed(tree.body):
+                if isinstance(node, ast.FunctionDef) and node.name == '__init__':
+                    instance_vars, _ = _pep224_docstrings(doc_obj, _init_tree=node)
+                    break
 
     try:
         ast_AnnAssign = ast.AnnAssign   # type: Type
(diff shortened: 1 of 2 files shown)


pdoc/__init__.py
```diff
@@ -262,10 +262,16 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
             tree = tree.body[0]  # ast.parse creates a dummy ast.Module wrapper
 
             # For classes, maybe add instance variables defined in __init__
-            for node in tree.body:
```
kernc, line 265: Ugh. Can't we make this just:
```py
for node in reversed(tree.body):
```

kernc (changes requested):
Add/amend a unit test. Maybe even this one:
https://github.com/pdoc3/pdoc/blob/6cfbc4fa0f09de79c51aedb435707a4dcd3b01b9/pdoc/test/__init__.py#L887

nosewings:
I've gone ahead and made the requested changes.

pdoc/test/__init__.py
```diff
@@ -889,22 +889,30 @@ def test_Variable_type_annotation_py36plus(self):
...
     var: int = 3
     """dummy"""
 
-    def __init__(self):
-        self.var2: float = 1
-        """dummy"""
+    @overload
+    def __init__(self, var2: float):
+        self.var2: float = var2
+        """dummy1"""
```
kernc, line 904: Maybe also change these two lines to `pass` so as to not be confusing as we don't really need/check them?

nosewings:
Okay, I've done that.

kernc (approved):

kernc:
Perfect, thanks!

PR #252, opened 2020-09-03
feat: Also scan modules with `inspect.ismodule` on public objects

Hywan:
When documenting an extension, (e.g. a `.so` file), `iter_modules`
can't be used to scan submodules because there is no files or
directories to iterate over. Nonetheless, the `<package>.__all__`
contains public objects of kind `ModuleType`. Simply by collecting
submodules to `Module.doc` with `inspect.ismodule` makes it possible
to generate documentation for submodules of an extension.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -20,7 +20,7 @@ import typing
 from contextlib import contextmanager
 from copy import copy
 from functools import lru_cache, reduce, partial
-from itertools import tee, groupby
+from itertools import tee, groupby, chain
 from types import ModuleType
 from typing import (
     cast, Any, Callable, Dict, Generator, Iterable, List, Mapping, Optional, Set, Tuple,
@@ -661,22 +661,27 @@ class Module(Doc):
                         if isdir(join(pth, file)) and '.' not in file:
                             yield file
 
-            for root in iter_modules(self.obj.__path__):
+            file_submodules = iter_modules(self.obj.__path__) if hasattr(self.obj, '__path__') else []
+            non_file_submodules = [obj for _, obj in public_objs if inspect.ismodule(obj)]
+
+            for root in chain(file_submodules, non_file_submodules):
                 # Ignore if this module was already doc'd.
                 if root in self.doc:
                     continue
 
+                root_name = root.__name__ if isinstance(root, ModuleType) else root
+
                 # Ignore if it isn't exported
-                if not _is_public(root) and not _is_whitelisted(root, self):
+                if not _is_public(root_name) and not _is_whitelisted(root_name, self):
                     continue
-                if _is_blacklisted(root, self):
-                    self._skipped_submodules.add(root)
+                if _is_blacklisted(root_name, self):
+                    self._skipped_submodules.add(root_name)
                     continue
 
                 assert self.refname == self.name
-                fullname = "%s.%s" % (self.name, root)
+
                 try:
-                    m = Module(import_module(fullname),
+                    m = Module(import_module(root if isinstance(root, ModuleType) else "%s.%s" % (self.name, root)),
                                docfilter=docfilter, supermodule=self,
                                context=self._context, skip_errors=skip_errors)
                 except Exception as ex:
@@ -811,9 +816,9 @@ class Module(Doc):
         """
         `True` if this module is a package.
 
-        Works by checking whether the module has a `__path__` attribute.
+        Works by checking `__package__`.
         """
-        return hasattr(self.obj, "__path__")
+        return not bool(self.obj.__package__)
 
     @property
     def is_namespace(self) -> bool:


kernc:
Is it possible to check those submodules later in:
https://github.com/pdoc3/pdoc/blob/6cfbc4fa0f09de79c51aedb435707a4dcd3b01b9/pdoc/__init__.py#L664
such as with:
```py
non_file_submodules = [name for name, obj in public_objs if inspect.ismodule(obj]

for root in chain(iter_modules(self.obj.__path__),
                  non_file_submodules):
    ...
```
This would need the extension module to pass `.is_package`:
https://github.com/pdoc3/pdoc/blob/6cfbc4fa0f09de79c51aedb435707a4dcd3b01b9/pdoc/__init__.py#L645
and its submodules the `import_module(...)` call:
https://github.com/pdoc3/pdoc/blob/6cfbc4fa0f09de79c51aedb435707a4dcd3b01b9/pdoc/__init__.py#L677-L679
Do they?

I'd prefer to keep submodule-loading code in one place if possible.

Hywan:
I'll try. I'm using https://github.com/PyO3/pyo3 (for the record).

Hywan:
> This would need the extension module to pass `.is_package`:

There is no `is_package` attribute on the submodule.

Hywan:
Well, it doesn't have a `__path__` attribute neither :-).

Hywan:
I'm not familiar with Python, but can't we use properties of `types.ModuleType`.

> `__package__`
>
> Which package a module belongs to. If the module is top-level (i.e. not a part of any specific package) then the attribute should be set to `''`, else it should be set to the name of the package (which can be `__name__` if the module is a package itself). Defaults to None.

We could rewrite `is_package` as follows:

```py
def is_package(self) -> bool:
    return not self.obj.__package__
```

Thoughts?

kernc:
If we have `a.b.c`, what does `b.__package__` hold? By the description could be either `'a'` (`b` belongs to package `a`) or `'a.b'` (`b` is a package itself).

Therefore:
```py
return self.obj.__package__ == self.obj.__name__
```
Sounds good if the tests pass.

Hywan:
In your example, `b.__package__` doesn't exist:

```py
>>> a.__package__ == ''
True
>>> a.b.__package__ == None
True
```

kernc:
> But `b` isn't a package, it's a submodule.

Even as it contains `a.b.c`? What is a package then?

I think this is where the current code was derived from: https://docs.python.org/3/glossary.html#term-package

Hywan:
With Pyo3, it is possible to create one package with submodules, but not subpackages. So if we have `a.b.c.`, `c` is a submodule of `b`, which is a submodule of `a`, which is a package.

My patch concerns submodules.

The think is that, at line 640:

```py
# If the module is a package, scan the directory for submodules
```

So first, you can scan the directory, but you can also scan the public object. And second, you should scan for submodules even if the current object is a module, not only if it is a package. The algorithm becomes generic then:

1. Scan the public objects and collect submodules
2. If the current object is a package, additionnally, collect more submodules by looking at the directory.

Thoughts?

kernc:
> you should scan for submodules even if the current object is a module, not only if it is a package

I agree we should check objects to see if they contain submodules. But according to [the definition of a package we will go by](https://docs.python.org/3/glossary.html#term-package): 

> A Python module which can contain submodules or, recursively, subpackages.

a module that contains submodules _is_ a package. So just make `Module.is_package` check pass for those Pyo3 objects. I'm not sure `not bool(self.obj.__package__)` will do it, though:

```py
>>> import numpy
>>> numpy.__package__
'numpy'
```

Hywan:
I confirm that `not bool(self.obj.__package__)` works for me!

pdoc/__init__.py
```diff
@@ -674,9 +677,9 @@ def iter_modules(paths):
                     continue
 
                 assert self.refname == self.name
-                fullname = "%s.%s" % (self.name, root)
+
                 try:
-                    m = Module(import_module(fullname),
+                    m = Module(import_module(root if isinstance(root, Module) else "%s.%s" % (self.name, root)),
```
Hywan, line 682: You may wonder why not using the `fullname` here? Because we can't import a submodule like `from a import b.c`, it is tied to the filesystem. For an extension, we can only use `from a import b; _ = b.c`. So if we try to `import_module` from a string, we will get an error.

Hywan:
Ah, wait a minute, it doesn't fully work. Continue to test.

Hywan:
OK now everything works like a charm for me :-).

kernc:
> I confirm that `not bool(self.obj.__package__)` works for me!

This absolutely won't work for anyone else as [the short example shows](https://github.com/pdoc3/pdoc/pull/252#issuecomment-687467855): numpy (correctly) defines `__package__` and it _is_ a package.

Hywan:
Why not removing `is_package` and simply using `inspect.ismodule(self.obj)`? Since a package is just a module, it's the `iter_modules` and public objects on `self.obj` that will look up for the submodules. Thoughts?

kernc:
`Module.is_package` is good semantic API and used in other places, so I motion to keep it. I also like the invariant `module.is_package == bool(module.submodules())`. Why not just make `.is_package` return True for those PyO3 modules?

pdoc/__init__.py
```diff
@@ -661,22 +661,27 @@ def iter_modules(paths):
...
-            for root in iter_modules(self.obj.__path__):
+            file_submodules = iter_modules(self.obj.__path__) if hasattr(self.obj, '__path__') else []
+            non_file_submodules = [obj for _, obj in public_objs if inspect.ismodule(obj)]
+
+            for root in chain(file_submodules, non_file_submodules):
                 # Ignore if this module was already doc'd.
                 if root in self.doc:
                     continue
 
+                root_name = root.__name__ if isinstance(root, ModuleType) else root
```
kernc, line 672: If you take `name` instead of `obj` in `non_file_submodules` above, you avoid this additional check. And the one in `import_module()` below.

Observing how:
```py
>>> import pdoc.cli    # Just a regular python package/module
>>> pdoc.cli.__name__  # Note, package prefix is *included*
'pdoc.cli'
>>> from pdoc import cli
>>> cli.__name__
'pdoc.cli'
```
I'm not entirely sure PyO3 isn't doing something wrong when building those modules. :thinking: Wouldn't be the first 'upstream' bug here. :sweat_smile:

pdoc/__init__.py
```diff
@@ -811,9 +816,9 @@ def is_package(self) -> bool:
         """
         `True` if this module is a package.
 
-        Works by checking whether the module has a `__path__` attribute.
+        Works by checking `__package__`.
         """
-        return hasattr(self.obj, "__path__")
+        return not bool(self.obj.__package__)
```
kernc, line 821: As a sole condition, this obviously won't work. Maybe as an additional condition. :thinking:

kernc: See what [the spec says about `__package__`](https://docs.python.org/3/reference/import.html#__package__) ...

pdoc/__init__.py
Hywan: We need `obj` rather than `name` because in `import_module` we need an object of kind `ModuleType` to avoid importing the module (i.e. to avoid calling `_module_path`) :-).

pdoc/__init__.py
kernc: > `from a import b.c`

That's invalid syntax! But one should be able to do either:
```py
from a.b import c  # or
import a.b.c
```
so a `fullname` string like `a.b.c` should work, no?

pdoc/__init__.py
kernc: I'm not convinced we need a module _object_ as opposed to its string representation. See https://github.com/pdoc3/pdoc/pull/252#discussion_r488609854.

Can you give me exact instructions on how to test this? What package do I install and try to document? What submodule do I look at?

Hywan: Thanks for your help.

If you follow those steps, https://github.com/wasmerio/wasmer-python#development, you should get a valid extension to test against. Once `rust` and `just` are installed, run the following commands:

```sh
$ just prelude
$ source .env/bin/activate
$ just build-all
$ just doc
```

Hywan: Or simpler, run:

```
$ just prelude
$ source .env/bin/activate
$ just build api
$ pdoc --html --output-dir docs/api --force wasmer
```

You don't need to generate all the packages/modules.

kernc: Got it, thanks. And what are the submodules that are not being picked up? What is the `a.b.c`?

Hywan: `wasmer.engine` or `wasmer.wasi`.

kernc: Can't for the life of me figure out how to `just build-all`. The command fails with:
```
just build compiler-llvm 
Build arguments: 
🐍 Found CPython 3.7m at python
   Compiling llvm-sys v100.2.0
   Compiling wasmer-engines-python v0.5.0 (/tmp/wasmer-python/packages/engines)
   Compiling inkwell v0.1.0-llvm10sample
 error: could not find native static library `Polly`, perhaps an -L flag is missing?
 
error: aborting due to previous error

error: could not compile `llvm-sys`.
```
`just build api`, however, worked. But this is what I get:
```py
>>> import wasmer                                                                          
>>> import wasmer.engine                                                                   
---------------------------------------------------------------------------
ModuleNotFoundError: No module named 'wasmer.engine'; 'wasmer' is not a package

>>> import wasmer.wasi                                                                     
---------------------------------------------------------------------------
ModuleNotFoundError: No module named 'wasmer.wasi'; 'wasmer' is not a package

>>> wasmer.__all__                                                                         
Out[162]: 
['__doc__',
 '__version__',
 '__core_version__',
 'Instance',
 'Module',
 'Value',
 'Memory',
 'Int16Array',
 'Int32Array',
 'Int8Array',
 'Uint16Array',
 'Uint32Array',
 'Uint8Array',
 'Buffer',
 'ExportKind',
 'ImportKind']
```
:thinking:

kernc:
I gave it a good think, and I resolved to being disinclined to list variable modules are proper submodules.

The fact to the matter is:
```py
>>> import wasmer                                                                          
>>> import wasmer.engine                                                                   
---------------------------------------------------------------------------
ModuleNotFoundError: No module named 'wasmer.engine'; 'wasmer' is not a package

>>> import wasmer.wasi                                                                     
---------------------------------------------------------------------------
ModuleNotFoundError: No module named 'wasmer.wasi'; 'wasmer' is not a package
```
Wasmer is not a package and therefore [does not have submodules](https://docs.python.org/3/glossary.html#term-package). It's `engine` and `wasi` would, therefore, be documented [merely as variables](https://github.com/pdoc3/pdoc/blob/d8b9dbc2a128099e89a9d7859720dac56d574094/pdoc/__init__.py#L645-L646), if PEP 224 docstring was present.

An analogous example is to have a module like:
```py
# a.py:

import re
__all__ = ['re']
```
and expect `re` to be listed as a submodule of `a`, even though `import a.re` would never work. :confused:

Thus, very strongly leaning to wontfix.

-------

I suggest you look into wasmer/PyO3 build configuration and see if you can make `wasmer` a proper package of its submodules. I'm not at all confident enough in my grasp of Python import machinery, but maybe even just declaring properties such as [`wasmer.__package__`](https://docs.python.org/3/reference/import.html#__package__) and [`wasmer.__path__`](https://docs.python.org/3/reference/import.html#__path__) (and likewise on the submodules) might do the trick ...

Issue #253, opened 2020-09-04
typing.NewType annotations rendered incorrectly


TristanCacqueray:
### Expected Behavior

The NewType name should be used in function signature.

### Actual Behavior

The NewType function repr is used.

### Steps to Reproduce

1. Define a NewType
2. Annotation a function with the new type
3. Render the documentation

### Additional info

- pdoc version: master

kernc:
For the record, I get this:
```py
>>> from typing import NewType
>>> newtype = NewType('MyType', str)
>>> inspect.formatannotation(newtype)
'<function NewType.<locals>.new_type at 0x7f48a9677400>'
```

TristanCacqueray:
And pdoc output looks like this:

```
$ cat demo.py
from typing import NewType

User = NewType('User', str)

def create(user: User) -> None:
    """create a user"""

$ pdoc3 demo.py
Module demo
===========

Functions
---------


`create(user: <function NewType.<locals>.new_type at 0x7f93d1ad0200>) -> NoneType`
:   create a user
```

Issue #231, opened 2020-07-03
nptyping annotations point to internal types


EmJay276:
Building the html documentation does not resolve the nptyping types correctly, see variable "C" below

### Expected Behavior

> <html lang="en">
> <body>
> <main>
> <article id="content">
> <section>
> <h2 class="section-title" id="header-classes">Classes</h2>
> <dl>
> <dt id="core.example.Material"><code class="flex name class">
> <span>class <span class="ident">Material</span></span>
> <span>(</span><span>E: float)</span>
> </code></dt>
> <dd>
> <div class="desc"><p>Stores material data</p></div>
> <h3>Instance variables</h3>
> <dl>
> <dt id="core.example.Material.C"><code class="name">var <span class="ident">C</span> : numpy.ndarray[(6, 6), Float[64]]</code> <-------------</dt>
> <dd>
> <div class="desc"><p>material stiffness matrix</p></div>
> </dd>
> <dt id="core.example.Material.E1"><code class="name">var <span class="ident">E1</span> : float</code></dt>
> <dd>
> <div class="desc"><p>Young's modulus</p></div>
> </dd>
> </dl>
> </dd>
> </dl>
> </section>
> </article>
> </body>
> </html>

### Actual Behavior

> <html lang="en">
> <body>
> <main>
> <article id="content">
> <section>
> <h2 class="section-title" id="header-classes">Classes</h2>
> <dl>
> <dt id="core.example.Material"><code class="flex name class">
> <span>class <span class="ident">Material</span></span>
> <span>(</span><span>E: float)</span>
> </code></dt>
> <dd>
> <div class="desc"><p>Stores material data</p></div>
> <h3>Instance variables</h3>
> <dl>
> <dt id="core.example.Material.C"><code class="name">var <span class="ident">C</span> : nptyping.types._ndarray.NDArray</code> <-------------</dt>
> <dd>
> <div class="desc"><p>material stiffness matrix</p></div>
> </dd>
> <dt id="core.example.Material.E1"><code class="name">var <span class="ident">E1</span> : float</code></dt>
> <dd>
> <div class="desc"><p>Young's modulus</p></div>
> </dd>
> </dl>
> </dd>
> </dl>
> </section>
> </article>
> </body>
> </html>

### Steps to Reproduce
```py
import nptyping as npt
import numpy as np


class Material:
    """
    Stores material data
    """
    E1: float
    C: npt.NDArray[(6, 6), np.double]

    def __init__(self, E: float):
[...]

PR #254, opened 2020-09-04
ENH: Better format typing.NewType annotations

TristanCacqueray:
Fixes: #253

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1166,6 +1166,20 @@ class Class(Doc):
         del self._super_members
 
 
+def _formatannotation(annot):
+    """
+    Format annotation, properly handling NewType types
+
+    >>> _formatannotation(NewType('MyType', str))
+    'MyType'
+    """
+    is_new_type = (getattr(annot, '__qualname__', '').startswith('NewType.') and
+                   annot.__module__ == 'typing')
+    if is_new_type:
+        return annot.__name__
+    return inspect.formatannotation(annot)
+
+
 class Function(Doc):
     """
     Representation of documentation for a function or method.
@@ -1264,7 +1278,7 @@ class Function(Doc):
         if isinstance(annot, str):
             s = annot
         else:
-            s = inspect.formatannotation(annot)
+            s = _formatannotation(annot)
             s = re.sub(r'\b(typing\.)?ForwardRef\((?P<quot>[\"\'])(?P<str>.*?)(?P=quot)\)',
                        r'\g<str>', s)
         s = s.replace(' ', '\N{NBSP}')  # Better line breaks in html signatures
@@ -1370,7 +1384,7 @@ class Function(Doc):
 
             formatted = p.name
             if p.annotation is not EMPTY:
-                annotation = inspect.formatannotation(p.annotation).replace(' ', '\N{NBSP}')
+                annotation = _formatannotation(p.annotation).replace(' ', '\N{NBSP}')
                 # "Eval" forward-declarations (typing string literals)
                 if isinstance(p.annotation, str):
                     annotation = annotation.strip("'")
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1174,7 +1174,7 @@ def _formatannotation(annot):
     'MyType'
     """
     is_new_type = (getattr(annot, '__qualname__', '').startswith('NewType.') and
-                   annot.__module__ == 'typing')
+                   getattr(annot, '__module__', '') == 'typing')
     if is_new_type:
         return annot.__name__
     return inspect.formatannotation(annot)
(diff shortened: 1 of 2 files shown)


pdoc/__init__.py
```diff
@@ -1166,6 +1166,18 @@ def _link_inheritance(self):
         del self._super_members
 
 
+def formatannotation(annot):
```
kernc, line 1169: Rename to `_formatannotation` as this will not be public API.

pdoc/test/example_pkg/_resolve_typing_forwardrefs/evaluated.py
```diff
@@ -1,5 +1,8 @@
-from typing import Set
+from typing import Set, NewType
+
+MyType = NewType('MyType', str)
```
kernc, line 3: Why touch this test also?

pdoc/__init__.py
```diff
@@ -1166,6 +1166,18 @@ def _link_inheritance(self):
...
+    """
+    Format annotation, properly handling NewType types
+
+    >>> formatannotation(NewType('MyType', str))
+    'MyType'
+    """
+    if callable(annot) and hasattr(annot, "__name__"):
+        return annot.__name__
```
kernc, line 1177: This is ill-posed if it's meant to match NewTypes only. In reality, this matches all functions, returning their names instead of reprs. I guess this is fine. But do amend the docstring.

kernc (changes requested):
Thanks for the PR! While you're reimplementing `formatannotation()`, can you maybe have a look if https://github.com/pdoc3/pdoc/issues/231 is an easy fix?

TristanCacqueray:
Unfortunately `typing.get_type_hints()` mentioned in #231 does not seems to resolve NewType functions and they are still represented as `<function NewType.<locals>.new_type at 0x414141>`

pdoc/__init__.py
TristanCacqueray: Done

pdoc/__init__.py
TristanCacqueray: Indeed, I guess this is why the py3.5 test are failing. Next commit rewrites the function to better check for NewTypes.

pdoc/test/example_pkg/_resolve_typing_forwardrefs/evaluated.py
TristanCacqueray: To add more coverage, well next commit removes that change.

pdoc/test/example_pkg/__init__.py
```diff
@@ -15,6 +16,12 @@
 
 __pdoc__ = {}
 
+MyType = NewType("MyType", str)
+
+
+def new_type(arg: MyType) -> None:
+    """NewType are correctly rendered"""
```
TristanCacqueray, line 23: Not sure where in the tests this should be asserted.

kernc: I think the addition to `test_Function_params()` (`bug253_newtype_annotation`) is enough as it covers and asserts. Can remove this hunk.

pdoc/__init__.py
```diff
@@ -1166,6 +1166,24 @@ def _link_inheritance(self):
...
 
 
+def _formatannotation(annot):
+    """
+    Format annotation, properly handling NewType types
+
+    >>> _formatannotation(NewType('MyType', str))
+    'MyType'
+    """
+    def is_new_type(annot):
```
kernc, line 1176: A variable suffices:
```py
is_newtype = (... and ...)
```

pdoc/test/example_pkg/__init__.py
TristanCacqueray: Removed.

pdoc/__init__.py
kernc: And the expression can be `getattr(annot, '__qualname__', '').startswith('NewType.') and annot.__module__ == 'typing'`.

TristanCacqueray: Thank you for the suggestion, looks good to me.

pdoc/__init__.py
```diff
@@ -1166,6 +1166,20 @@ def _link_inheritance(self):
...
 
+def _formatannotation(annot):
+    """
+    Format annotation, properly handling NewType types
+
+    >>> _formatannotation(NewType('MyType', str))
+    'MyType'
+    """
+    is_new_type = (getattr(annot, '__qualname__', '').startswith('NewType.') and
+                   annot.__module__ == 'typing')
```
kernc, line 1177:
```suggestion
                   getattr(annot, '__module__', '') == 'typing')
```

kernc (approved):

kernc:
Lgtm, thanks!

PR #274, opened 2020-10-18
ENH: Add support for reST-style docstrings

### LINKED ISSUE #110: Feature Request: Support reStructuredText-style docstrings

### LINKED ISSUE #273: No support for classic Sphinx Style docstring format?
I have a project that uses "classic" style docstrings, that style/format gets sometimes called "Sphinx style" and sometimes "reST" style:

```
"""
Some text to describe what the function does. 
This can span several lines.

:param param1: description of first parameter.
   can also span several lines if indented.
:param param2: another one
:returns: description of what is returned
"""
```

This will not detect the `:param` sections and format them properly in the generated HTML.

Am I doing this wrong or is this not supported?
In general, is it possible to somehow configure the expected docstring format or is everything supposed to get
auto-detected (so different files or even different parts in a file could have different formats)? 

I am using pdoc version 0.9.1 with Python 3.7

devdrian:
Hi kernc and all contributors,

Addressing issues #110 and #273, I added support for reST-style docstrings, following your suggestions in #198.

Additionally, in case the docformat is not explicitly specified, it will be distinguished between reST-style and Numpy-/Google-style in `html_helpers.to_markdown`.

Fixes #110 
Fixes #273 
Closes #198

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -12,6 +12,11 @@ from functools import partial, lru_cache
 from typing import Callable, Match
 from warnings import warn
 import xml.etree.ElementTree as etree
+import docutils.nodes
+import docutils.core
+import collections
+from typing import *
+
 
 import markdown
 from markdown.inlinepatterns import InlineProcessor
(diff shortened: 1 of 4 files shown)


pdoc/html_helpers.py
```diff
@@ -265,6 +265,71 @@ def googledoc_sections(match):
...
+        # the sections
+        parameter_types = {}
+        return_types = []
+        for tag, name, body in regex.findall(text):
+            if tag == 'type':
+                parameter_types[name] = body.strip()
+            elif tag == 'rtype':
+                return_types.append(body.strip())
+
+        active_section = None  # Keep track of the currently active section (e.g. Args, Returns)
```
kernc, line 328: I don't think I like this state machine too much. Why not simply:
```py
for tag, name, body in regex.findall(text):
    ...  # fill in relevant info dicts

... # serialize dicts in some predefined, expected order
```

pdoc/test/example_pkg/__init__.py
```diff
@@ -298,6 +298,25 @@ def doctests(self):
...
         """
 
+    def reST(self):
+        """
+        Summary line.
+
+        :param arg1: Text1
+        :parameter arg2: Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed
+            diam nonumy eirmod tempor invidunt
+        :arg arg_arg_3: Y:=H^T*X!@#$%^&&*()_[]{}';'::
```
kernc, line 308: Lovely tests. But mind the three-part param spec: [`:param int n: 42`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists).

pdoc/test/example_pkg/__init__.py
```diff
@@ -298,6 +298,25 @@ def doctests(self):
...
+        :type arg1: int
+        :type arg2: Optional[List[Tuple[str]]]
+        :type arg_arg_3: Dict[int, Dict[str, Any]]
+        :return: True. Or False. Depends
+        :rtype: bool
```
kernc, line 313: Also, then, [`var`, `ivar`, `cvar`, and `vartype`, and `meta`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists).

pdoc/html_helpers.py
```diff
@@ -462,6 +543,9 @@ def to_markdown(text: str, *,
         if 'numpy' in docformat:
             text = _ToMarkdown.numpy(text)
 
+        if 'rest' in docformat:
```
kernc, line 546: Python people say this is [`'restructuredtext'`](https://github.com/search?l=Python&q=org%3APython+__docformat__&type=Code).

kernc (changes requested):
I like it. But now I worry people will expect all sorts of further reST magic to work, such as references (``:class:`foo.Foo` ``; `:obj:`, `:func:`, `:module:`, ...), anchors and hyperlinks ...

I wonder if we shouldn't instead incorporate docutils, or maybe defer to `docutils.publish_string()` at least the descriptions. :thinking:

pdoc/html_helpers.py
```diff
@@ -430,20 +496,35 @@ def to_markdown(text: str, *,
...
     example template.
     """
     if not docformat:
-        docformat = str(getattr(getattr(module, 'obj', None), '__docformat__', 'numpy,google '))
+        docformat = str(getattr(getattr(module, 'obj', None), '__docformat__', ''))
+
+        # Infer docformat if it hasn't been specified
+        if docformat == '':
+            reST_tags = ['param', 'arg', 'type', 'raise', 'except', 'return', 'rtype']
+            reST_regex = fr'^:(?:{"|".join(reST_tags)}).*?:'
```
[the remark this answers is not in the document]
muecelee: Good point, I added a space to the tags where it's required.

pdoc/html_helpers.py
```diff
@@ -430,20 +496,35 @@ def to_markdown(text: str, *,
...
+            found_reST_tags = re.findall(reST_regex, text, re.MULTILINE)
+
+            # Assume reST-style docstring if any of the above specified tags is present at the beginning of a line.
+            # Could make this more robust, e.g., by checking against the amount of found google or numpy tags
+            if len(found_reST_tags) > 0:
```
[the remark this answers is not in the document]
muecelee: Done.

pdoc/test/example_pkg/__init__.py
muecelee: Thanks. I added some more tests, also for general reST syntax

pdoc/test/example_pkg/__init__.py
muecelee: Are now processed and tested for

pdoc/html_helpers.py
muecelee: Changed it. Also changed in `test_docformat()` the docformat to "epytext" instead of "restructuredtext", since it's now processed.

pdoc/html_helpers.py
muecelee: Yeah, it definitely was a bit rough and dirty. But I changed it now for the new version and followed your approach.

muecelee:
Thanks @kernc for your review and your suggestions!

I changed it now to utilize docutils. It's not the cleanest solution right now: Everything but the tag lists - for which Markdown is still  generated - is converted directly to html using docutils' publish functions. But it definitely opens up the possibility in the future to integrate some more advanced features.

For reST I also disabled the admonitions preprocessing, since docutils processes them directly. However, some of the directives from the reST_directives test are not supported and generate a warning message. I chose to leave the messages - which are placed in a div at the corresponding position - in for now, since it allows the user to fix this. But we can discuss if that's the best way. For the reST test, I adapted the admonitions to only include ones that are directly supported by docutils.

collerek:
Hi @kernc any update on merging this PR? ;)
Something is still not resolved?

Ljancek:
Any update?

makseq:
@kernc @collerek Why don't you merge it? :slightly_frowning_face:

kenahoo:
I'd love for `:params:` support to be merged, anything we can do to help test @kernc ?

scottzach1:
Hi @kernc, I would love to see this functionality within pdoc3.

Is there anyway I can help progress this? 😊

kenahoo:
The `pdoc` code appears to have not been touched in several years.  It might be time to either fork it or find an alternative.

Issue #270, opened 2020-10-12
Python 3.5 "end of life"


kernc:
With [Python 3.5.x having reached its end of life/support](https://www.python.org/downloads/release/python-3510/), we can begin to use "advanced" features from Python 3.6+:

* [f-strings](https://www.python.org/dev/peps/pep-0498/)
* ???

tnwei:
f-strings is the only thing that comes to mind as the significant feature added in Python 3.6 that is relevant to `pdoc`. From going through the project codebase, I've noticed few inconsistencies in string formatting (single quotes, double quotes, %s substitution, .format), so moving to f-strings is definitely a great opportunity to standardize that. 

Just want to throw the question though, how will moving the supported base version of Python from 3.5+ to 3.6+ impact `pdoc`'s users? Updating all docs to use f-strings would be more of a formatting change rather than a feature added that augments fuctionality. Genuine question as I'm not familiar w/ the user base.

If we do move ahead w/ bumping to Python 3.6+ and standardizing string formatting to using f-strings, I would gladly submit a PR for that.

kernc:
> Just want to throw the question though, how will moving the supported base version of Python from 3.5+ to 3.6+ impact `pdoc`'s users? Updating all docs to use f-strings would be more of a formatting change rather than a feature added that augments fuctionality. Genuine question as I'm not familiar w/ the user base.

Neither am I, really. :smiley: Debian Buster with Python 3.7 was released over a year ago, and Python 3.6 was already in Ubuntu 18.04 LTS. CI platforms such as Travis and GH Actions take Python version as a parameter, regardless of distro. So even as far as _stable_ and _late_ environments go, few users should be affected IMO. F-strings and related changes would just count as refactoring towards an improved and future-compatible codebase. The first version to feature those changes would then mark a point of incompatibility with <Py3.6, with Py3.5 users happily (:crossed_fingers:) stuck on whatever version released by then (say, 9.2 or so). If it later turns out there are some grave fixes to cater to a large important group of excluded users, we can run a backport branch, but I really don't count on it.

PR welcome. :+1:

tnwei:
take

tnwei:
Submitted a PR! 

Should there be a code style section added to `CONTRIBUTING.md` to specify formatting preference to keep things neat moving ahead?

kernc:
> Should there be a code style section added to CONTRIBUTING.md to specify formatting preference to keep things neat moving ahead?

[...]

PR #278, opened 2020-10-23
Standardize string formatting to use f-strings

tnwei:
As per #270 , submitting this pull request to revamp string formatting in `pdoc` to use f-strings. Also bumped min requirements of Python 3.5+ to Python 3.6+ in `setup.py`.

Following `CONTRIBUTING.md`:
+ All tests passed from `python setup.py test`
+ Ran `flake8 --max-line-length 100`, no new warnings. At time of writing, existing warning is:
```
 ./pdoc/test/example_pkg/_resolve_typing_forwardrefs/postponed.py:1:1: F407 future feature annotations is not defined
```


pdoc/__init__.py
```diff
@@ -776,7 +776,7 @@ def _link_inheritance(self):
                 continue
             if not isinstance(docstring, str):
                 raise ValueError('__pdoc__ dict values must be strings;'
-                                 '__pdoc__[{!r}] is of type {}'.format(name, type(docstring)))
+                                 f'__pdoc__[{name!r}] is of type {type(docstring)}')
```
kernc, line 779: Missing space at the end of first line.

pdoc/cli.py
```diff
@@ -236,8 +236,8 @@ def do_GET(self):
                 import traceback
                 from html import escape
                 code = 404
-                out = "Error importing module <code>{}</code>:\n\n<pre>{}</pre>".format(
-                    self.import_path_from_req_url, escape(traceback.format_exc()))
+                out = f"Error importing module <code>{self.import_path_from_req_url}</code>:"
+                f"\n\n<pre>{escape(traceback.format_exc())}</pre>"
```
kernc, line 240: Indent, wrap in parentheses.

pdoc/__init__.py
```diff
@@ -140,9 +140,10 @@ def _render_template(template_name, **kwargs):
         t = tpl_lookup.get_template(template_name)
     except TopLevelLookupException:
         raise OSError(
-            "No template found at any of: {}".format(
-                ', '.join(path.join(p, template_name.lstrip("/"))
-                          for p in tpl_lookup.directories)))
+            "No template found at any of: "
+            f"{', '.join(path.join(p, template_name.lstrip('/')) for p in tpl_lookup.directories)}"
```
kernc, line 144: Maybe just `"one" + "other"` or a new `paths` variable?

kernc (commented):
Note, _pdoc/html_helpers.py_ and _pdoc/templates/{html,pdf}.mako_ contain some further `str.format()` / `str.__mod__` calls.

pdoc/__init__.py
tnwei: Can you please elaborate where is the missing space? I've checked that the before/after edit prints the same line

pdoc/cli.py
tnwei: Something like this looks better?
``` python
out = (
    "Error importing module "
    f"<code>{self.import_path_from_req_url}</code>:\n\n"
    f"<pre>{escape(traceback.format_exc())}</pre>"
)
```

pdoc/__init__.py
tnwei: Yeah that looks better, how about the following?
```python
paths = (path.join(p, template_name.lstrip('/')) for p in tpl_lookup.directories)
raise OSError(f"No template found at any of: {', '.join(paths)}")

**Comment by tnwei** [comment]
> Note, pdoc/html_helpers.py and pdoc/templates/{html,pdf}.mako contain some further str.format() / str.__mod__ calls.

You are right, will be checking on the missed files

File: pdoc/__init__.py (hunk 1 above)
Comment (kernc) [reply pdoc/__init__.py:779]: > ```py
> '__pdoc__ dict values must be strings; '
> ```

Was missing before already. :sweat_smile:

Comment (tnwei) [reply pdoc/__init__.py:779]: Ah, got it.

**Comment by tnwei** [comment]
Re-checked all documents to pick up on missed `__mod__` and `.format()` calls:
+ [X] `pdoc/__init__.py`
+ [X] pdoc/cli.py
+ [X] `pdoc/test/__init__.py`
+ [ ] pdoc/html_helpers.py (line 176)
+ [X] pdoc/templates/html.mako
+ [ ] pdoc/templates/pdf.mako (line 18)

**Comment by tnwei** [comment]
Two questions:

#### pdoc/html_helpers.py (line 176)
`str.format` takes output from a `map` nested in list comprehension as below:
``` python
return '\n\n'.join('`{}`\n:   {}'.format(*map(str.strip, line.split(':', 1)))
                       for line in filter(None, spec_with_desc.split('\n')))
```

Reformatting it would give
``` python
return '\n\n'.join(f'`{line.split(":", 1)[0].strip()}`\n:   '
            f'{line.split(":", 1)[1].strip()}' for line in filter(None, spec_with_desc.split('\n')))
```
but it looks rather inelegant. Are we good with this?

#### pdoc/templates/pdf.mako (line 18)

``` python
text = re.sub(r'^(?P<indent>( *))!!! \w+ \"([^\"]*)\"(.*(?:\n(?P=indent) +.*)*)',
                      lambda m: '{}**{}:** {}'.format(m.group(2), m.group(3),
                                                      re.sub('\n {,4}', '\n', m.group(4))),
                      text, flags=re.MULTILINE)
```

One of the inputs to `str.format` above include evaluating regex substitution where newline chars are involved. f-strings don't allow backslash in the curly braces, and assigning it to a variable will need rewriting the lambda function that is passed to `re.sub()`, like so:
``` python
def paragraph_fmt(m):
    sub = re.sub('\n {,4}', '\n', m.group(4))
    return f'{m.group(2)}**{m.group(3)}:** {sub}'
text = re.sub(r'^(?P<indent>( *))!!! \w+ \"([^\"]*)\"(.*(?:\n(?P=indent) +.*)*)',
                     paragraph_fmt, text, flags=re.MULTILINE)
```

Would love to have your comments @kernc

kernc:
Snippets like these show `str.format/%` still has some value. And how awful regexes are.

I don't mind new variables in both cases. In the first case maybe:
```py
spec, desc = map(...)
return f"..."
```
When in doubt, opt for clarity.

tnwei:
> Snippets like these show str.format/% still has some value. And how awful regexes are.

Thanks for the thoughts, at some point I was contemplating if it's better to leave them as-is actually. 

Anywho, I've finished updating each file in their own commit, whole PR is ready for review.

pdoc/html_helpers.py
```diff
@@ -277,13 +280,13 @@ def _admonition(match, module=None, limit_types=None):
...
         if type in ('image', 'figure'):
-            return '{}![{}]({})\n'.format(
-                indent, text.translate(str.maketrans({'\n': ' ',
-                                                      '[': '\\[',
-                                                      ']': '\\]'})).strip(), value)
+            strfmt = text.translate(str.maketrans({
+                '\n': ' ',
+                '[': '\\[',
+                ']': '\\]'})).strip()
+            return f'{indent}![{strfmt}]({value})\n'
```
kernc, line 289:
```suggestion
            alt_text = text.translate(str.maketrans({
                '\n': ' ',
                '[': '\\[',
                ']': '\\]'})).strip()
            return f'{indent}![{alt_text}]({value})\n'
```

kernc (approved):

kernc:
Many thanks @tnwei! Hope you got your t(r)ee. :smile:

Issue #291, opened 2020-12-19
setup.py test doesn't seem to work


csm10495:
### Expected Behavior

```
python setup.py test
```

should work as shown here: https://github.com/pdoc3/pdoc/blob/master/CONTRIBUTING.md

### Actual Behavior

Python 3.8.5 on Windows, commit: bec63077a03b2

```
C:\Users\csm10495\Desktop\pdoc>python setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
running egg_info
writing pdoc3.egg-info\PKG-INFO
writing dependency_links to pdoc3.egg-info\dependency_links.txt
writing entry points to pdoc3.egg-info\entry_points.txt
writing requirements to pdoc3.egg-info\requires.txt
writing top-level names to pdoc3.egg-info\top_level.txt
reading manifest template 'MANIFEST.in'
writing manifest file 'pdoc3.egg-info\SOURCES.txt'
running build_ext
Traceback (most recent call last):
  File "setup.py", line 15, in <module>
    setup(
  File "C:\Python38\lib\site-packages\setuptools\__init__.py", line 163, in setup
    return distutils.core.setup(**attrs)
  File "C:\Python38\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python38\lib\distutils\dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:\Python38\lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "C:\Python38\lib\site-packages\setuptools\command\test.py", line 238, in run
    self.run_tests()
  File "C:\Python38\lib\site-packages\setuptools\command\test.py", line 256, in run_tests
    test = unittest.main(
  File "C:\Python38\lib\unittest\main.py", line 100, in __init__
    self.parseArgs(argv)
  File "C:\Python38\lib\unittest\main.py", line 147, in parseArgs
    self.createTests()
  File "C:\Python38\lib\unittest\main.py", line 158, in createTests
    self.test = self.testLoader.loadTestsFromNames(self.testNames,
  File "C:\Python38\lib\unittest\loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "C:\Python38\lib\unittest\loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
[...]

Issue #287, opened 2020-12-14
Preserve Whitespace?


csm10495:
Is there a flag/setting to preserve Whitespace within multiline docstrings? Right now my multiline docstrings appear to have their newlines lost in translation to html. 

Thanks.

kernc:
HTML doesn't really care about newlines. Currently no provisions to replace `'\n'` with `'<br>'` (i.e. line breaks) for HTML output.

Can you show an example docstring that doesn't look right?

csm10495:
Sure: https://www.github.com/csm10495/pynom/tree/master/pynom%2Fpynom.py look for the __init__ of PyNom. 

Then the docs are here: https://csm10495.github.io/pynom/pynom.html

The newlines weren't preserved in the docstrings which made them look kind of weird compared to the original source code.

kernc:
Well, there's no indication anywhere in the docs that pdoc should preserve newlines. Pdoc3 supports Numpydoc and Google docstring formats as well as raw Markdown, but no `<pre>`.

If you care to format your docstrings a little more nicely, following e.g. the [Google style guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) which results in similar to what you already have, they should render better.

Alternatively, you're welcome to edit your [_html.mako_ template](https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates) and wrap docstrings in `<pre>` tags.

PR #292, opened 2020-12-19
Add support for variable doc-comments starting with '#:'

### LINKED ISSUE #289: Docstrings for variables from doc-comments above
I see that 

```
module_variable = 1
"""Docstring for module_variable."""
```

works for giving the module_variable a docstring.

Is there anyway that something like this could also (or optionally) be supported: 

```
# Docstring for module_variable.
module_variable = 1
```

It seems rather rare for a docstring to follow after a variable declaration; for me it's much more commonly in front of the declaration. It also is more common to be of the single line-comment (via #) variety. 

Similar ask for class variable docstrings as well.

csm10495:
I'd like to write a unit-test for this functionality, though I'm having some trouble running the existing suite: #291 

This basically matches up with sphinx behavior and a similar implementation.

Fixes #289
Resolves https://github.com/pdoc3/pdoc/issues/291

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -244,7 +244,7 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                        _init_tree=None) -> Tuple[Dict[str, str],
                                                  Dict[str, str]]:
     """
-    Extracts PEP-224 and '#:' docstrings for variables of `doc_obj`
+    Extracts PEP-224 docstrings and doc-comments (`#: ...`) for variables of `doc_obj`
     (either a `pdoc.Module` or `pdoc.Class`).
 
     Returns a tuple of two dicts mapping variable names to their docstrings.
@@ -305,6 +305,9 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                   target.value.id == 'self'):
                 name = target.attr
 
+        if name and not _is_public(name) and not _is_whitelisted(name, doc_obj):
+            name = None
+
         return name
 
     # For handling PEP-224 docstrings for variables
@@ -316,9 +319,6 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
 
         name = get_name(assign_node)
         if not name:
-            continue
-
-        if not _is_public(name) and not _is_whitelisted(name, doc_obj):
             continue
 
         docstring = inspect.cleandoc(str_node.value.s).strip()
@@ -336,13 +336,8 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
         if not name:
             continue
 
-        if not _is_public(name) and not _is_whitelisted(name, doc_obj):
-            continue
-
         def get_line_indentation(line):
-            # There isn't a definitive definition for tab width, but lets just be consistent.
-            line = line.replace('\t', '    ')
-            return len(line) - len(line.lstrip(' '))
+            return len(line) - len(line.lstrip())
 
         if name not in vars:
             # There wasn't a PEP-224 style docstring, so look for a '#:' style one above
@@ -365,7 +360,8 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                 comment_lines.append(assignment_line.rsplit('#:', 1)[-1])
 
             if comment_lines:
-                vars[name] = '\n\n'.join(comment_lines)
+                # Adding 2 spaces to the end of each line to ensure we keep line breaks
+                vars[name] = '\n'.join([c + '  ' for c in comment_lines])
 
     return vars, instance_vars
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -347,17 +347,17 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
             comment_lines = []
             for line in reversed(source_lines[:assign_node.lineno - 1]):
                 if get_line_indentation(line) == assignment_line_indentation and \
-                        line.lstrip().startswith('#:'):
-                    comment_lines.append(line.split('#:', 1)[1])
+                        line.lstrip().startswith('#: '):
+                    comment_lines.append(line.split('#: ', 1)[1])
                 else:
                     break
 
             # Since we went 'up' need to reverse lines to be in correct order
             comment_lines = comment_lines[::-1]
 
-            # Finally: check for a '#:' comment at the end of the assignment line itself.
-            if '#:' in assignment_line:
-                comment_lines.append(assignment_line.rsplit('#:', 1)[-1])
+            # Finally: check for a '#: ' comment at the end of the assignment line itself.
+            if '#: ' in assignment_line:
+                comment_lines.append(assignment_line.rsplit('#: ', 1)[-1])
 
             if comment_lines:
                 # Adding 2 spaces to the end of each line to ensure we keep line breaks
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -360,8 +360,7 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                 comment_lines.append(assignment_line.rsplit('#: ', 1)[-1])
 
             if comment_lines:
-                # Adding 2 spaces to the end of each line to ensure we keep line breaks
-                vars[name] = '\n'.join([c + '  ' for c in comment_lines])
+                vars[name] = '\n'.join(comment_lines)
 
     return vars, instance_vars
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -285,9 +285,6 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                     break
 
     def get_name(assign_node):
-        target = None
-        name = None
-
         if isinstance(assign_node, ast.Assign) and len(assign_node.targets) == 1:
             target = assign_node.targets[0]
         elif isinstance(assign_node, ast.AnnAssign):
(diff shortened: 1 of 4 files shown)


pdoc/__init__.py
```diff
@@ -319,6 +327,46 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+                    comment_lines.append(line.split('#:', 1)[1])
+                else:
+                    break
+
+            # Since we went 'up' need to reverse lines to be in correct order
+            comment_lines = comment_lines[::-1]
+
+            # Finally: check for a '#:' comment at the end of the assignment line itself.
+            if '#:' in assignment_line:
+                comment_lines.append(assignment_line.rsplit('#:', 1)[-1])
```
kernc, line 365: Interestingly, this fails on a contrived line like:
```py
MARKER = '#:'  #: Doc-comment marker
```
Is this how Sphinx does it?

pdoc/__init__.py
```diff
@@ -319,6 +327,46 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+
+            if comment_lines:
+                vars[name] = '\n\n'.join(comment_lines)
```
kernc, line 368: I think this should be a single `\n`, otherwise it leaves blank lines between consecutive comment lines ...
```py
#: This long-winded text that stretches over multiple lines
#: should likely become a single documentation paragraph.
variable = 1
```
If you want to force line breaks (i.e. https://github.com/pdoc3/pdoc/issues/287), there are [ways for that](https://commonmark.org/help/tutorial/03-paragraphs.html) in markdown.

kernc (changes requested):
Looks neat! But definitely will need a unit test. :grimacing:

pdoc/__init__.py
csm10495: Swapped to using 2 spaces at the end of each line.

pdoc/__init__.py
```diff
@@ -319,6 +327,46 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+        if not name:
+            continue
+
+        if not _is_public(name) and not _is_whitelisted(name, doc_obj):
+            continue
+
+        def get_line_indentation(line):
+            # There isn't a definitive definition for tab width, but lets just be consistent.
+            line = line.replace('\t', '    ')
+            return len(line) - len(line.lstrip(' '))
```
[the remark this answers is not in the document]
csm10495: Yep. You're right. I was still thinking of Python 2 where I had seen (obnoxious usage of tabs with spaces for indentation). Not needed in a Python 3 world.

pdoc/__init__.py
```diff
@@ -319,6 +327,46 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+    for assign_node in ast.iter_child_nodes(tree):
+        if not isinstance(assign_node, (ast.Assign, ast.AnnAssign)):
+            continue
+
+        name = get_name(assign_node)
+        if not name:
+            continue
+
+        if not _is_public(name) and not _is_whitelisted(name, doc_obj):
+            continue
```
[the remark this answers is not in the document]
csm10495: Sure thing.

pdoc/__init__.py
```diff
@@ -244,7 +244,7 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
                        _init_tree=None) -> Tuple[Dict[str, str],
                                                  Dict[str, str]]:
     """
-    Extracts PEP-224 docstrings for variables of `doc_obj`
+    Extracts PEP-224 and '#:' docstrings for variables of `doc_obj`
```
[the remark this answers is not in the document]
csm10495: 👍

csm10495:
@kernc, added unit tests, fixed running on Windows. Let me know if it needs other changes. Otherwise feel free to merge.

csm10495:
@kernc let me know if further changes are needed. Thanks!

pdoc/test/__init__.py
```diff
@@ -1661,5 +1660,98 @@ def test_head(self):
             self.assertEqual(cm.exception.code, 304)
 
 
+class InlineVariableDocstringsTest(unittest.TestCase):
```
kernc, line 1663: In like with current tests structure, I'd prefer all this merged into one single test case:
```py
#: Not included

#: Line 1
#: Line 2
var1 = 1  #: Line 3

#: Not included
var2 = 1
"""PEP-224 takes preference"""

class C
    #: class var
    class_var = 1
    
    def __init__(self):
       #: instance var
       self.instance_var = 1
```
e.g. in `pdoc.test.example_pkg._doccomments`, loaded simply like this test:
https://github.com/pdoc3/pdoc/blob/bec63077a03b2136e6982825d327073ed76cc20a/pdoc/test/__init__.py#L792-L798
and with its five or so assertions.

pdoc/test/__init__.py
```diff
@@ -106,25 +106,21 @@ class CliTest(unittest.TestCase):
...
+        os.path.join('example_pkg', 'subpkg2'),
+        os.path.join('example_pkg', 'subpkg2', '_private.html'),
+        os.path.join('example_pkg', 'subpkg2', 'module.html'),
+        os.path.join('example_pkg', 'subpkg2', 'index.html'),
     ]
-    PUBLIC_FILES = [f for f in ALL_FILES if '/_' not in f]
-
-    if os.name == 'nt':
-        ALL_FILES = [i.replace('/', '\\') for i in ALL_FILES]
-        PUBLIC_FILES = [i.replace('/', '\\') for i in PUBLIC_FILES]
```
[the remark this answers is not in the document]
csm10495: Nope, unfortunately a couple tests were failing.  Some other places were using / so I updated here and in those places.

pdoc/test/__init__.py
csm10495: That seems odd to me since if something is wrong, you don't immediately know what it is.

The way the tests I wrote are: if one piece (or case) fails you know based off the name, what happened. When throwing it all in one, you lose that granularity; you just know something is wrong.

Though since you prefer, I can shove it all in one case with one py file tested.

pdoc/__init__.py
```diff
@@ -319,6 +327,42 @@ def _pep224_docstrings(doc_obj: Union['Module', 'Class'], *,
...
+            # Since we went 'up' need to reverse lines to be in correct order
+            comment_lines = comment_lines[::-1]
+
+            # Finally: check for a '#: ' comment at the end of the assignment line itself.
+            if '#: ' in assignment_line:
+                comment_lines.append(assignment_line.rsplit('#: ', 1)[-1])
+
+            if comment_lines:
+                # Adding 2 spaces to the end of each line to ensure we keep line breaks
+                vars[name] = '\n'.join([c + '  ' for c in comment_lines])
```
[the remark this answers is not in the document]
csm10495: I disagree with this for my own usage, though if that's the desire: Ok. Will change.

pdoc/test/__init__.py
kernc: I know, but it's not such a major component to warrant a separate test class. The question I see relevant to cover is: _do the doc-comment docstrings work correctly or not_?

pdoc/__init__.py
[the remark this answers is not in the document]
kernc: I understand, issue https://github.com/pdoc3/pdoc/issues/287. Your way makes it impossible to produce documentation paragraphs _without_ line breaks.
```py
#: I don't want to force new lines in prose documentation just
#: because I opted for line breaks in e.g. 69-column source.
```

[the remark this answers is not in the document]
csm10495: Fair enough. You're right. In theory also adding a \ at the end would force a break so that help too.

pdoc/test/__init__.py
csm10495: Ok. Will do

csm10495:
@kernc I've made the requested changes. Let me know if it seems ok now. Thanks!

pdoc/test/__init__.py
```diff
@@ -1099,6 +1098,44 @@ def func(self):
...
+
+            cls_name_to_docs, instance_name_to_docs = pdoc._pep224_docstrings(pdoc_module)
+            self.assertEqual(cls_name_to_docs, {
+                                                'var2': 'PEP-224 takes precedence',
+                                                'var1': 'Line 1\nLine 2\nLine 3'})
+
+            pdoc_class = pdoc.Class('C', pdoc_module, module.C)
+            cls_name_to_docs, instance_name_to_docs = pdoc._pep224_docstrings(pdoc_class)
+            self.assertEqual(cls_name_to_docs, {'class_var': 'class var'})
+            self.assertEqual(instance_name_to_docs, {'instance_var': 'instance var'})
```
kernc, line 1137:
```suggestion
            mod = pdoc.Module(filename)
            self.assertEqual(mod.doc['var1'].docstring, 'Line 1\nLine 2\nLine 3')
            self.assertEqual(mod.doc['var2'].docstring, 'PEP-224 takes precedence')
            self.assertEqual(mod.doc['C'].docstring, '')
            self.assertEqual(mod.doc['C'].doc['class_var'].docstring, 'class var')
            self.assertEqual(mod.doc['C'].doc['instance_var'].docstring, 'instance var')
```

Let's use public API to cover, not the implementation details. Ad lib, maybe this passes.

kernc:
Now with this feature in place, amendment of documentation is in order:
https://github.com/pdoc3/pdoc/blob/28b366a57ae475168d5039d94470cda925066075/pdoc/documentation.md#L106-L137

pdoc/test/__init__.py
csm10495: Ok. Done.

csm10495:
@kernc added documentation, fixed test. Please check again. Thanks.

kernc:
Applied with slight modification.

This is a great feature. Many thanks @csm10495! :sparkles:

kernc (approved):

PR #300, opened 2021-01-06
When using --http, also reload submodules - fix #264

### LINKED ISSUE #264: HTTP Server doesn't detect changes if __all__ is used
### Expected Behavior
I expect changes to be reflected when using the web server when I make changes to files, even though I use `__all__ `and view the module.

### Actual Behavior
Changes are not reflected.

### Steps to Reproduce

1. Create folder `my_module`

2. Add two files
```
#### __init__.py

from .file1 import my_func

__all__ = ['my_func',]
```

```
#### file1.py

def my_func(x: int):
    """

    Parameters
    ----------
    x : int
        This is a number

    Returns
    -------
    int
        Same as input

    """
    return x
```

3. Run `pdoc --http localhost:1234 my_module` from parent dir.

4. Go to `localhost:1234/my_module`

5. Change `file1-py`  to:
```
#### file1.py

def my_func(x: int):
    """

    Parameters
    ----------
    x : int
        This is a number

    Returns
    -------
    int
        Very different output

    """
    return x
```

6. Notice no change on the web page

### Additional info

- pdoc version: 0.9.1

Lucas-C:
Fixes #264 

The fix in `cli.py` was made to avoid this error while pressing `F5` in my browser when using `--http`:

```
Error importing module fpdf:

  File "/home/lucas_cimon/.local/share/virtualenvs/pyfpdf/lib/python3.7/site-packages/pdoc/cli.py", line 234, in do_GET
    out = self.html()
  File "/home/lucas_cimon/.local/share/virtualenvs/pyfpdf/lib/python3.7/site-packages/pdoc/cli.py", line 267, in html
    skip_errors=args.skip_errors,
AttributeError: 'Namespace' object has no attribute 'skip_errors'
```

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -230,6 +230,10 @@ def import_module(module: Union[str, ModuleType],
     # `isinstance(..., pdoc.Doc)` calls won't work correctly.
     if reload and not module.__name__.startswith(__name__):
         module = importlib.reload(module)
+        # We recursively reload all submodules, in case __all_ is used - cf. issue #264
+        for mod_key, mod in list(sys.modules.items()):
+            if mod_key.startswith(module.__name__):
+                importlib.reload(mod)
     return module
 
 
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -265,7 +265,7 @@ class _WebDoc(BaseHTTPRequestHandler):
         """
         return pdoc.html(self.import_path_from_req_url,
                          reload=True, http_server=True, external_links=True,
-                         skip_errors=args.skip_errors,
+                         skip_errors=self.args.skip_errors,
                          **self.template_config)
 
     def resolve_ext(self, import_path):


kernc:
I was wondering how come _I_ hadn't thought of that simple solution. Maybe it has some drawbacks I'm no longer aware of, but it seems to work. :smiley: 

Much appreciated!

kernc:
Ah, right, I remember now what I was thinking. This fix only applies when imported objects are from submodules of the current module, which might be a common, but certainly not the general case.

In terms of #264, the fix "breaks" if observed `my_func()` were defined in `__init__` and imported/exposed in `file1`.

Ideas?

Lucas-C:
Maybe #264 can be re-opened to track this.
In my case this fix was enough to le me edit the docs interactively.

I may have an idea for a fix though:
instead of checking `if mod_key.startswith(module.__name__)`,
we could check `if mod_key.startswith(base_mod)`
where `base_mod` is the root module of `module.__name__`
(extracted with `module.__name__.split(".", 1)[0]` for example).

What do you think?

kernc:
It's not too bad an idea at all. This will then work for all modules within a single package. :+1:

PR #302, opened 2021-01-18
Fix a fatal error preventing documentation generation in some cases

mn3mos:
Got a fatal exception on this line, which prevented the generation of the documentation.

The problem is that some variable is initialized in a try bloc but still used out of it afterwards.
If the getattr fails line 682 - which was my problem - then the variable is used without being initialized, which lead to fatal error.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -683,12 +683,13 @@ class Module(Doc):
             for name in self.obj.__all__:
                 try:
                     obj = getattr(self.obj, name)
-                    if not _is_blacklisted(name, self):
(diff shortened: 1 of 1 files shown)


kernc:
> If the getattr fails line 682 - which was my problem - then the variable is used without being initialized

That should be caught by `AttributeError` or what do you mean variable fails? Can you construct a mwe example that reproduces the issue? This was just recently touched in https://github.com/pdoc3/pdoc/commit/9bdb076e267b33949fb6ffc906316e8651d5f474.

Issue #268, opened 2020-10-12
Local image not copied over to output directory


fr34q:
### Expected Behavior

When adding a local image (relative path) to the Python Docstring like this:
```python
"""
# Test

.. image:: test.svg
"""
```
then the image should be shown in the generated documentation of this Python file.

### Actual Behavior

No image is shown, although the HTML element is generated:
```html
<img alt="" src="test.svg">
```
The network traffic (when hosted by `pdoc3 --http`) shows a 302 response with redirection to `test.svg/` which then directly afterwards returns a 404 Not Found. The pdoc log states:
```log
127.0.0.1 - - [12/Oct/2020 09:43:32] "GET /pdt/test.html HTTP/1.1" 200 -
127.0.0.1 - - [12/Oct/2020 09:43:32] "GET /pdt/test.svg HTTP/1.1" 302 -
127.0.0.1 - - [12/Oct/2020 09:43:32] "GET /pdt/test.svg/ HTTP/1.1" 404 -
```
Remote images (using URL) work but local images do not, which is a shame when the idea is to put all documentation within the repository/code. I also tested different image formats like png and jpg, but this did not make a difference.

### Steps to Reproduce

1. Create folder `/tmp/pdoctest`, within, create a subfolder `pdt`
2. In the subfolder, copy an image file, create an empty `__init__.py` and a `test.py` file
3. Paste the above docstring in the `test.py` file and save it. Make sure that the name of the image matches.
4. Run `PYTHONPATH="/tmp/pdoctest" pdoc3 --http : pdt`
5. In your browser, open `http://localhost:8080/pdt/test.html`
6. The heading "Test" should be visible but not the image

### Additional info

- pdoc version: 0.8.1

If you require further information, please let me know!

fr34q:
I just noticed that someone else already posted this issue here - sorry for this:
https://github.com/pdoc3/pdoc/issues/267

I will add a comment there :)

PR #304, opened 2021-01-22
Serve images via http server

### LINKED ISSUE #267: HTTP server will not serve images
### Expected Behavior
Images are served by http server

### Actual Behavior
Images return a 302 and a URL that ultimately results in a 404.

### Steps to Reproduce
1. Generate documentation that has images
2. Run the http server

### Additional info
```$ pdoc modelgen modelgen.modelgen --html --output-dir doc --force --http :
Starting pdoc server on localhost:8080
pdoc server ready at http://localhost:8080
127.0.0.1 - - [08/Oct/2020 19:05:00] "GET /modelgen.modelgen/ HTTP/1.1" 200 -
127.0.0.1 - - [08/Oct/2020 19:05:00] "GET /modelgen.modelgen/img/screw_hole.svg HTTP/1.1" 302 -
127.0.0.1 - - [08/Oct/2020 19:05:00] "GET /modelgen.modelgen/img/screw_hole.svg/ HTTP/1.1" 404 -
```

Note that accessing the files directly via `file://` works as expected.

- pdoc version: 0.9.1

gunnarvoet:
Both png and jpg/jpeg files are now shown in the generated docs when using the `--http` option. Fixes #267.

Adds two local images to `exampl_pkg`.


pdoc/cli.py
```diff
@@ -222,7 +223,17 @@ def do_GET(self):
                     out = f"External identifier <code>{import_path}</code> not found."
             else:
                 return self.redirect(resolved)
-        # Redirect '/pdoc' to '/pdoc/' so that relative links work
```
kernc, line 225: This comment line belongs to the section beneath. Let's leave it in.

pdoc/cli.py
```diff
@@ -222,7 +223,17 @@ def do_GET(self):
...
+        # Deal with images
```
kernc, line 226: Did you maybe investigate `SimpleHTTPRequestHandler.do_GET()` as [advised in the issue](https://github.com/pdoc3/pdoc/issues/267#issuecomment-706210873)?
I'm fairly confident something like this would work:
```
import mimetypes
from http.server import SimpleHTTPRequestHandler

...

# More robust, includes e.g. SVGs
is_image = mimetypes.guess_type(self.path).startswith('image/')

...

elif is_image:
    return SimpleHTTPRequestHandler.do_GET(self)

...
```

pdoc/cli.py
gunnarvoet: I put this back in.

pdoc/cli.py
gunnarvoet: I added the mime type guessing in the new commit. I'm not sure what the advantage of adding `SimpleHTTPRequestHandler.do_GET()` would be?

kernc: One advantage seems not having to maintain all these lines below (response code, headers, file IOErrors ...)?

pdoc/test/example_pkg/__init__.py
```diff
@@ -313,10 +313,20 @@ def reST_directives(self):
...
+        Remote image:
+
         .. image:: https://www.debian.org/logos/openlogo-nd-100.png
 
-        Now you know.
+        Local image png format (should be a cropped Debian logo):
+
+        .. image:: ./images/local_logo.png
+        
+        Local image jpg format (should be a cropped Debian logo):
```
kernc, line 326: Can then use a single example image since we defer to mimetype.

pdoc/cli.py
gunnarvoet: Tried `SimpleHTTPRequestHandler.do_GET(self)` but couldn't get it to work. It complains about the `_WebDoc` object missing the attribute `send_head`.

pdoc/test/example_pkg/__init__.py
gunnarvoet: Removed the second test image in the new commit.

pdoc/cli.py
kernc: Ah. And what if `_WebDoc` is made to inherit `SimpleHttpRequestHandler`? :grimacing:

pdoc/test/example_pkg/__init__.py
kernc: I think you forgot the image file itself.

gunnarvoet: Removed both image files and added one that clearly shows that it is a local file.

pdoc/cli.py
gunnarvoet: Changed `_WebDoc` to inherit `SimpleHTTPRequestHandler` and it seems to work.

kernc: Great, I really like this better. And it's more generic in case additional filetypes are excepted in the future.

pdoc/cli.py
```diff
@@ -243,10 +256,11 @@ def do_GET(self):
                 )
                 out = out.replace('\n', '<br>')
 
-        self.send_response(code)
-        self.send_header("Content-type", "text/html; charset=utf-8")
-        self.end_headers()
-        self.echo(out)
+        if not is_image:
```
kernc, line 259: Since we return early, this check is no longer relevant.

pdoc/test/example_pkg/__init__.py
```diff
@@ -313,10 +313,16 @@ def reST_directives(self):
...
+        Now you know.
+
+        Remote image:
+
         .. image:: https://www.debian.org/logos/openlogo-nd-100.png
 
-        Now you know.
+        Local image:
 
+        .. image:: ./images/local_image.png
```
kernc, line 324: As I understand, this doesn't touch https://github.com/pdoc3/pdoc/issues/268, so this image actually won't be visible in [pdoc web docs](https://pdoc3.github.io/pdoc/doc/pdoc/test/example_pkg/#pdoc.test.example_pkg.reST_directives), instead showing a 404 box. I wonder if we need to test it so thoroughly. :thinking:

pdoc/cli.py
gunnarvoet: Agreed - I removed this.

pdoc/cli.py
gunnarvoet: Yes, works like magic now. Thanks for the tip.

pdoc/test/example_pkg/__init__.py
gunnarvoet: I have almost hacked a pull request together that should solve #268. It is nice to have the image for testing. In fact, I am now also linking to the image from `subpkg` and `subpkg2`.

pdoc/test/example_pkg/subpkg/images
```diff
@@ -1 +1 @@
-../images
\ No newline at end of file
+../images/
```
gunnarvoet, line 1: Seems like symlinks can not be followed? How do the existing links in `subpkg` and `subpkg2` work though?

kernc: I'm not sure. But I must admit I'm kind of disinclined to hosting that image. Wouldn't a test in `pdoc.test.HttpTest`, such as:
```py
with NamedTemporaryFile(suffix='.png') as image_file:
    dirname, basename = os.path.split(image_file.name)
    with self._http([dirname]) as url:
       ...
```
be much simpler, and more reliable?

gunnarvoet: I don't care too much whether the image is hosted or not, but don't have the time to figure out how to write a test as suggested above.

PR #309, opened 2021-02-11
Filter out "dummy" when listing for class content

pums974:
In the module page, the hidden inherited methods appears as "dummy" class variable:
image

The proposed patch, simply removes them:
image

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -72,6 +72,7 @@ if os.getenv("XDG_CONFIG_HOME"):
 # If you know how to keep the warning, but skip the object creation
 # altogether, please make it happen!
 class _BLACKLISTED_DUMMY:
+    """dummy"""
     pass
 
 
@@ -703,7 +704,7 @@ class Module(Doc):
 
         for name, obj in public_objs:
             if obj is _BLACKLISTED_DUMMY:
-                self.doc[name] = Variable(name, self, '__pdoc__dummy', obj=obj)
+                self.doc[name] = Variable(name, self, 'dummy', obj=obj)
             elif _is_function(obj):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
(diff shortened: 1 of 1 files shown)


kernc:
Huh? Where does this come from? :flushed: 

The only place we're setting "dummy" is in the test your patch fails:
https://github.com/pdoc3/pdoc/blob/49b7773ba6a96ddee63bf5c4cd48df250ab12c36/pdoc/test/__init__.py#L940-L945

pums974:
No there is also line 706
https://github.com/pdoc3/pdoc/blob/4b94c42838f2dbcb043ce8382be310c682623aaa/pdoc/__init__.py#L704-L708
and line 1050
https://github.com/pdoc3/pdoc/blob/4b94c42838f2dbcb043ce8382be310c682623aaa/pdoc/__init__.py#L1048-L1052

In my case it's the second one that shows up

pdoc/__init__.py
```diff
@@ -1143,7 +1143,9 @@ def _filter_doc_objs(self, type: Type[T], include_inherited=True,
                          filter_func: Callable[[T], bool] = lambda x: True,
                          sort=True) -> List[T]:
         result = [obj for obj in _filter_type(type, self.doc)
-                  if (include_inherited or not obj.inherits) and filter_func(obj)]
+                  if (include_inherited or not obj.inherits)
+                  and filter_func(obj)
+                  and obj.docstring != "__pdoc__dummy"]
```
kernc, line 1148: Could we, rather than filter by docstring, filter by:
```py
obj.obj is _BLACKLISTED_DUMMY
```
which should be the case.

This way, we avoid [stringly types](https://wiki.c2.com/?StringlyTyped).

Might even move that literal `'dummy'` string into `_BLACKLISTED_DUMMY.__doc__` (class docstring) to avoid future confusion. :sweat_smile:

pums974: Okay, I'll update the PR next week

kernc:
@pums974 Is there a simple test case we can include/[extend](https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/test/__init__.py#L611-L632) as a regression test for this issue?

pums974:
I don't know, probably.

But I'm really not an expert on this, and I have to admit that I have trouble understanding your tests.

PR #318, opened 2021-03-01
Recurse through submodules in extension modules

### LINKED ISSUE #319: Submodules not showing up for (native) extension modules
When running `pdoc` on an extension modules (aka, a native "C" extensions), the extension module's submodules don't show up in the documentation even though &lt;TAB>-autocomplete in a Python REPL can find the submodules. This seems to be because `pdoc` searches for submodules by inspecting the source directory, which isn't available for extension modules.

I've proposed PR #318 to fix this issue. The proposed solution works but I'm not sure if it is safe enough to remove the old "source directory traversal" method. I'd appreciate guidance on completing the PR.

### Expected Behavior

Running `pdoc` on a native extension module should generate documentation for the entire extension module, including its submodules.

### Actual Behavior

- With current master: only items in the top-level module appear in the documentation. Submodules don't show up in the documentation.
- With the proposed fix in #318: works as expected.

### Steps to Reproduce

The following steps generate a minimalistic native extension module in Rust that exhibits the problem. The language shouldn't matter though.

1. Install a rust toolchain, see https://rustup.rs
2. Create the following directory structure:
```text
pyext/
├── Cargo.toml
└── src/
    └── lib.rs
```
with the following file contents:
    - `Cargo.toml`:
```toml
[package]
authors = ["Name <em@i.l>"]
edition = "2018"
name = "pyext"
version = "0.1.0"

[lib]
crate-type = ["cdylib"]

[dependencies]
pyo3 = {version = "0.13.2", features = ["extension-module"]}
```
    - `src/lib.rs`:
```rust
use pyo3::{prelude::*, wrap_pymodule};

/// Docstring of main module.
#[pymodule(pyext)]
fn init_main_module(_py: Python<'_>, module: &PyModule) -> PyResult<()> {
    module.add_wrapped(wrap_pymodule!(submodule))?;
    Ok(())
}

/// Docstring of submodule
#[pymodule(submodule)]
fn init_submodule(_py: Python<'_>, submodule: &PyModule) -> PyResult<()> {
    submodule.add("variable", 42)?;
    Ok(())
}
```
3. Compile the extension module: `cargo build`
4. Create a properly named symlink to the object file:
    - on Linux: `ln -s target/debug/libpyext.so pyext.so`
    - on Mac: `ln -s target/debug/libpyext.dylib pyext.so`
    - on Windows: rename `target\debug\libpyext.dll` to `pyext.pyd`
5. Start a Python REPL from the directory containing the `pyext.so` file and verify that the submodule exists and can be found by tab completion:
```text
$ python
Python 3.6.10 (default, May 22 2020, 17:59:48) 
[GCC 9.2.1 20191008] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyext
>>> pyext.<TAB>  --> autocompletes to "pyext.submodule", proving that the submodule can be found
>>> pyext.submodule.variable
42
```
6. Run `pdoc --html pyext` from the same directory.
    - With the version at current master, the generated documentation leaves out the submodule.
    - With the version proposed in #318, the generated documentation includes the submodule.

### Additional info

- pdoc version: master and 0.9.2 don't work, the one from #318 works.
- tested on linux

robamler:
Fixes https://github.com/pdoc3/pdoc/issues/319

The original method of finding submodules by recursing through the
directory structure didn't work with extension modules (aka native "C"
extensions) because there is no directory structure to recurse through.
This commit adds code to recurse through submodules by inspecting
`obj.__all__`, i.e., using the same mechanism that is already used for
finding variables, functions, and classes.

This is a minimalistic first sketch of a solution. It works on a native
extension that I tried it on but it introduces some duplication since
there are now two strategies to search for submodules. For a native
module, the first strategy (searching the directory structure) will fail
but the second (new) strategy (using `obj.__all__`) will work. For
regular modules, both strategies will likely work and so the second
strategy will just overwrite the results from the first strategy with
identical objects. This should technically work but it seems
unnecessary. I just don't feel confident enough to remove the first
strategy without further guidance.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -708,6 +708,10 @@ class Module(Doc):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
                 self.doc[name] = Class(name, self, obj)
+            elif inspect.ismodule(obj):
+                self.doc[name] = Module(
+                    obj, docfilter=docfilter, supermodule=self,
+                    context=context, skip_errors=skip_errors)
             elif name in var_docstrings:
                 self.doc[name] = Variable(name, self, var_docstrings[name], obj=obj)
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -708,10 +708,6 @@ class Module(Doc):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
                 self.doc[name] = Class(name, self, obj)
-            elif inspect.ismodule(obj):
-                self.doc[name] = Module(
-                    obj, docfilter=docfilter, supermodule=self,
-                    context=context, skip_errors=skip_errors)
             elif name in var_docstrings:
                 self.doc[name] = Variable(name, self, var_docstrings[name], obj=obj)
 
@@ -765,6 +761,24 @@ class Module(Doc):
                 if m.is_namespace and not m.doc:
                     del self.doc[root]
                     self._context.pop(m.refname)
+        elif hasattr(self.obj, '__all__'):
+            # Python extension modules don't get recognized by `is_package` because they have no
+            # "__path__" attribute. We treat them here separately. We support submodules of
+            # extension modules only if they are explicitly exposed via the "__all__" attribute
+            # because otherwise it's hard to distinguish proper submodules from re-exports (i.e.,
+            # the function  `is_from_this_module` doesn't work on submodules).
+            for name, obj in public_objs:
+                if inspect.ismodule(obj) and not hasattr(obj, '__file__') and name not in self.doc:
+                    try:
+                        m = Module(
+                            obj, docfilter=docfilter, supermodule=self,
+                            context=self._context, skip_errors=skip_errors)
+                    except Exception as ex:
+                        if skip_errors:
+                            warn(str(ex), Module.ImportWarning)
+                            continue
+                        raise
+                    self.doc[name] = m
 
         # Apply docfilter
         if docfilter:
--- a/pdoc/test/__init__.py
+++ b/pdoc/test/__init__.py
@@ -472,6 +472,35 @@ class ApiTest(unittest.TestCase):
                     self.assertEqual(sorted(m.name for m in m.submodules()),
                                      [EXAMPLE_MODULE + '.' + m for m in submodules])
 
+    @ignore_warnings
+    def test_module_without_path(self):
+        # GH-319: https://github.com/pdoc3/pdoc/issues/319
+        parent_module = ModuleType('parent_module')
+        child_module1 = ModuleType('child_module1')
+        child_module2 = ModuleType('child_module2')
+        grandchild_module = ModuleType('grandchild_module')
+
+        child_module1.grandchild_module = grandchild_module
+        child_module1.__all__ = ['grandchild_module']
+
+        parent_module.child_module1 = child_module1
+        parent_module.child_module2 = child_module2
+        parent_module.__all__ = ['child_module1', 'child_module2']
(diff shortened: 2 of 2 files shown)


pdoc/__init__.py
```diff
@@ -708,6 +708,10 @@ def is_from_this_module(obj):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
                 self.doc[name] = Class(name, self, obj)
+            elif inspect.ismodule(obj):
+                self.doc[name] = Module(
+                    obj, docfilter=docfilter, supermodule=self,
+                    context=context, skip_errors=skip_errors)
```
kernc, line 714:
```suggestion
                    context=self._context, skip_errors=skip_errors)
```

kernc (commented):
Can you construct a minimal, self-contained test with a module that has no path, but contains a reference to another module, and assert both modules are discovered? Like:
https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/test/__init__.py#L42


> For regular modules, the second strategy will just overwrite the results

The procedure actually skips modules that have the same name as one that was already defined:
https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/__init__.py#L734-L737
Though this method seems unreliable.

Further, if I:
```py
import .foo as bar
```
module `foo` will now be be documented as `foo` and as `bar` ... :thinking:

pdoc/__init__.py
```diff
@@ -708,6 +708,10 @@ def is_from_this_module(obj):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
                 self.doc[name] = Class(name, self, obj)
+            elif inspect.ismodule(obj):
```
kernc, line 711: Could we maybe move this down into `if self.is_package`-`else` block? I suppose built extension modules don't match the condition in `Module.is_package()`? That way only one of the two discovery paths will run.

pdoc/__init__.py
robamler: Done.

pdoc/__init__.py
robamler: Thanks for the guidance, I did that in f379105.

robamler:
Thanks for the quick feedback! I've moved the new code into the `if self.is_package - else` block and added a test.

Studying the existing code for "normal" submodule discovery, I realized I should be more careful to avoid capturing cycles and reexports (not sure if they're possible in native extension modules though). So I made sure the new code only runs if the parent module explicitly exposes public members through the `__all__` attribute. This restriction seems to be necessary because the [`is_from_this_module` function](https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/__init__.py#L690-L692) doesn't work on submodules (`inspect.getmodule(x)` returns `x` rather than `x`'s parent if `x` is a submodule). This restriction doesn't seem to be a problem for the native extension modules I tried, but I'd be happy to relax it if you have an idea how.

> Further, if I:
> 
> ```python
> import .foo as bar
> ```
> 
> module `foo` will now be be documented as `foo` and as `bar` ... :thinking: 

Is that possible with native extension modules? I know too little about Python's module system to tell.

robamler:
One problem I encountered: the HTML documentation for a native extension module now places all files in the top level directory. The contents of the files shows the correct nested hierarchy of submodules but the paths for the corresponding HTML files doesn't reflect this, so paths may clash if, e.g., a submodule has the same name as its parent module. Any idea how to fix this?

Issue #324, opened 2021-03-19
Optional[Type] is shown as Union[Type, NoneType]


hhoppe:
In the doc I maintain, https://google.github.io/mediapy/, the functions have many `typing` attributes.
These generally look great in the `pdoc3` output, except that `Optional[Type]` is replaced by the more confusing `Union[Type, NoneType]`.
Perhaps it would be possible to recognize this pattern and somehow change it back?

hhoppe:
Also, the return type `-> None` is shown using the lengthier `-> NoneType`.

kernc:
> except that `Optional[Type]` is replaced by the more confusing `Union[Type, NoneType]`

I'm afraid that's what the values are resolved as by _typing_ internals:
```py
>>> import typing
>>> def foo() -> typing.Optional[int]: pass
>>> typing.get_type_hints(foo)
{'return': typing.Union[int, NoneType]}
```

If you can think of a good matching/replacement scheme, this would be the place to override it:
https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/__init__.py#L1243-L1258

hhoppe:
Thanks for the pointers.  Handling the general case looks difficult.  It seems that inspect uses `repr` to create a string from an annotation, which can be arbitrarily nested.  One could try to recognize the pattern(s) in the resulting string, but that would require a parser rather than a regular expression.

PR #325, opened 2021-03-20
Replace Union[T, NoneType] in all type annotations by Optional[T].

hhoppe:
This addresses https://github.com/pdoc3/pdoc/issues/324

Also, uniformly removes prefix "typing." from type annotations, and replaces "NoneType" by "None",
so that the function signatures look more similar to those commonly found in source code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
         run: |
           find -name '*.md' | xargs .github/lint-markdown.sh
           flake8
-          mypy pdoc
+          mypy -p pdoc
           time coverage run -m unittest -v pdoc.test
           PDOC_TEST_PANDOC=1 time catchsegv python -m unittest -v pdoc.test.CliTest.test_pdf_pandoc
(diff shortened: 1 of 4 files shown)


hhoppe:
@kernc Unfortunately it looks like we worked in parallel... :-)   Interesting to compare the solutions and see many similarities.  Please have a look to see if my efforts have useful ideas.  Thanks.

hhoppe:
The changes in my very last commit "Fix for flake8" https://github.com/pdoc3/pdoc/pull/325/commits/d0101475868d52cf24aa69976891931011adbbae is wrong.  Sorry about that.

I did a checkout of https://github.com/pdoc3/pdoc/commit/6bd00e13a2a5839ba0b20afa8b4f6321164942a2 and its output exactly reproduces what I had generated, so I'm happy with the changes in the main branch.

kernc:
Sincere apologies. I went to investigate, and then I had half of it working and couldn't just throw it away. I also realized even existing `nptyping.` and `NewType` replacements didn't trigger when nested, so something had to be done.
Thanks for verifying.

PR #336, opened 2021-04-16
Mention the obj property which can be assigned additional properties

### LINKED ISSUE #335: Custom attributes to HTML templates
I would like to append the documentation with auxiliary data which would be collected at the time of document generation. (eg. test results).

### Expected Behavior

When using `pdoc3` programmatically, it would be nice if additional variables could be attached to the Class/Method objects, which could be accessed from within the templates.

### Actual Behavior

Currently I'm unable to find any way to add custom variables.

### Additional info

- pdoc version: `0.9.2`

silverbucket:
Fixes #335

### DIFF
--- a/pdoc/documentation.md
+++ b/pdoc/documentation.md
@@ -312,7 +312,7 @@ For example:
 
 .. tip::
     To add custom properties that can be accessible from the template,
-    `pdoc.Class.obj` points to the underlying type object with which additional
+    `pdoc.Class.obj` points to the underlying `type` object with which additional
     properties can be added. `Module` and `Function` likewise.
 
 When documenting a single module, you might find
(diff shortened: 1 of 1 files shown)


silverbucket:
@kernc can we merge this?

kernc:
@silverbucket Apologies for the long turnaround. I follow the saying, when in doubt, defer. 😅 I understand you found this workaround when you [weren't able to assign new attributes to pdoc documentation objects directly](https://github.com/pdoc3/pdoc/issues/335#issuecomment-820801809) due to our use of [`__slots__`](https://docs.python.org/3/reference/datamodel.html#slots)?
https://github.com/pdoc3/pdoc/blob/65342aecd782326b4c98492bbf9102c3534a735a/pdoc/__init__.py#L520
In that case, with Python becoming leaner in the recent years, I'm questioning the (ancient) use of `__slots__` and would probably prefer to simply remove the line ...

Issue #152, opened 2020-01-29
Darkmode


jappi00:
Hey,

I think we all love Darkmodes. What about pdoc? I think a default integrated Darkmode would be great! What do you think about something like this? Thank you for this awesome package!

kernc:
We do. :smile:

Integrated, but optional behind a [config toggle](https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates), would indeed make sense. Or even something purely JS-based that shows a light switch on the client side for the user to switch at will. :thinking: :bulb: 

I won't be able to work on this shortly, so further discussion / PR welcome!

kernc:
Or, if the way of a config tunable, a dict of preferred style colors would empower the user more. Something like:
```py
# config.mako:

bright_color_theme = dict(foreground='black',
                          background='white',
                          link='...',
                          ...)
dark_color_theme = ...
color_theme = bright_color_theme  # Default
```

kernc:
An even better idea is to provide a basic set of easily overridable colors in CSS variables (e.g. in _pdoc/templates/colors.css_) like it's currently for highlighting color:
https://github.com/pdoc3/pdoc/blob/cd84fb3b0c512a203742cfdfaa26ba6fb0e535aa/pdoc/templates/css.mako#L6-L8
https://github.com/pdoc3/pdoc/blob/cd84fb3b0c512a203742cfdfaa26ba6fb0e535aa/pdoc/templates/css.mako#L184-L186

xcodz-dot:
Hello, I was also looking for dark theme for pdoc3, can you please tell what fields am i supposed to modify in css file?

kernc:
@xcodz-dot it's not available yet. There's just [the idea](https://github.com/pdoc3/pdoc/issues/152#issuecomment-677903920) to separate colors hard-coded all over _css.mako_ into another file _colors.css_ (or maybe _theme.css_), with CSS variables defined on the _:root_ pseudo-class.

That new file then needs to be included in the _html.mako_ template:
https://github.com/pdoc3/pdoc/blob/aef1917a668ca9f3cd26373d1d69bcb5001108cc/pdoc/templates/html.mako#L389-L392

[...]

PR #341, opened 2021-05-12
WIP: Added theming and dark mode support per #152.

pm5k:
## Abstract
As per https://github.com/pdoc3/pdoc/issues/152 -- added support for easy theming and dark-mode.

## Changelog

- Created a `themes` folder inside of `templates/` so there's a place for any number of user or maintainer-generated theme files.
- Created a `default.mako` theme alongside of a blank `custom.mako` theme. The user should probably use the custom file if they wish to add their own theme, or alternatively they can copy the default one, rename it and adjust settings therein.
- Moved all color defs from `css.mako` to `default.mako` so these can be built by the template renderer independently.
- 'css.mako' now only contains non-color styles and imports theme from 'default.mako' or whichever template the user specifies.
- Enabled automatic dark-mode switching based on user's system preference. This is handled by javascript inside of `html.mako`.
- Enabled switching modes using a button toggle on the docs themselves, which overrides system prefs.
- Enabled saving toggled theme state in local storage for convenience.
- Changed the `hljs` config parameter to now be a tuple of strings, where the first is `light-mode` and second is `dark-mode`, this gets loaded accordingly to the previously mentioned toggles to keep hljs blocks consistent with overall theme.
- Created "dark theme" based on Github's own dark mode. It now lives in `default.mako`.
- Widened `.content` as it was wasting real-estate.
- Adjusted `hljs` code blocks as the extra padding was not needed.
- Re-jigged media queries, as there's no need to split mobile/desktop. So I simply added one breakpoint for 1024.
- Set default "portables" breakpoint as 1024px, to better reflect modern-day screen sizes.

## Demo from issue comment

pdoc3_dm_demo_01 (1)
pdoc3_dm_demo_02

### DIFF
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -11,7 +11,7 @@
         return name
     url = dobj.url(relative_to=module, link_prefix=link_prefix,
                    top_ancestor=not show_inherited_members)
-    return f'<a title="{dobj.refname}" href="{url}">{name}</a>'
+    return '<a title="{}" href="{}">{}</a>'.format(dobj.refname, url, name)
 
 
   def to_html(text):
@@ -285,7 +285,16 @@
   supermodule = module.supermodule
   %>
   <nav id="sidebar">
+    <div id="dark-mode-toggle" onclick="toggleDarkMode()">
+      <svg class="dm-icon-off" viewBox="0 0 16 16">
+        <path d="M7 0h2v2H7zM12.88 1.637l1.414 1.415-1.415 1.413-1.414-1.414zM14 7h2v2h-2zM12.95 14.433l-1.415-1.414 1.414-1.414 1.415 1.413zM7 14h2v2H7zM2.98 14.363L1.566 12.95l1.415-1.414 1.414 1.415zM0 7h2v2H0zM3.05 1.707L4.465 3.12 3.05 4.535 1.636 3.121z" />
+        <path d="M8 4C5.8 4 4 5.8 4 8s1.8 4 4 4 4-1.8 4-4-1.8-4-4-4z" />
+      </svg>
 
+      <svg class="dm-icon-on" viewBox="0 0 16 16">
+        <path d="M6,0C2.5,0.9,0,4.1,0,7.9C0,12.4,3.6,16,8.1,16c3.8,0,6.9-2.5,7.9-6C9.9,11.7,4.3,6.1,6,0z"></path>
+      </svg>
+    </div>
     <%include file="logo.mako"/>
 
     % if google_search_query:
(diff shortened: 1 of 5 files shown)


pm5k:
@kernc

Do let me know what your thoughts are regarding this. Additionally, following contribution guidelines, I ran flake8 and unittest, while flake8 came up with nothing, unit test had one failing test even before my change:

```bash
  File "/home/tom/pdoc/pdoc/test/__init__.py", line 100, in wrapper
    func(*args, **kwargs)
  File "/home/tom/pdoc/pdoc/test/__init__.py", line 1039, in test_Class_params
    self.assertEqual(pdoc.Class('G2', mod, G2).params(), ['*args', '**kwds'])
AssertionError: Lists differ: [] != ['*args', '**kwds']

Second list contains 2 additional elements.
First extra element 0:
'*args'

- []
+ ['*args', '**kwds']
```
Unsure if relevant, but thought I'd mention.

Would also be good to know if more changes are needed and get general feedback on this.

Thanks

pdoc/templates/html.mako
```diff
@@ -434,7 +517,7 @@
 
 <footer id="footer">
     <%include file="credits.mako"/>
-    <p>Generated by <a href="https://pdoc3.github.io/pdoc" title="pdoc: Python API documentation generator"><cite>pdoc</cite> ${pdoc.__version__}</a>.</p>
+    <p>Generated by <a href="https://pdoc3.github.io/pdoc"><cite>pdoc</cite> ${pdoc.__version__}</a>.</p>
```
pm5k, line 520: Will have to revert this, the copy-over must've blasted it..

pdoc/templates/html.mako
```diff
@@ -11,7 +11,7 @@
         return name
     url = dobj.url(relative_to=module, link_prefix=link_prefix,
                    top_ancestor=not show_inherited_members)
-    return f'<a title="{dobj.refname}" href="{url}">{name}</a>'
+    return '<a title="{}" href="{}">{}</a>'.format(dobj.refname, url, name)
```
pm5k, line 14: Will have to revert this, the copy-over must've blasted it..

pdoc/templates/html.mako
```diff
@@ -383,13 +392,87 @@
...
+
+  <%namespace name="css" file="css.mako"/>
+  <style id="css-theme">${css.load_theme()}</style>
+  <script>
+    const html = document.querySelector("html");
+    const isDarkQuery = window.matchMedia("(prefers-color-scheme: dark)");
+
+    const changeHLJS = (style) => {
+      const hljsStyle = document.querySelector("#hljs");
+      const styleURL = "https://cdnjs.cloudflare.com/ajax/libs/highlight.js/10.1.1/styles/" + style + ".min.css";
```
pm5k, line 407: Could probably do with making this a parameter/config value tbh..

davfsa:
@pm5k Just wanted to drop in and say, amazing work. It looks splendid :heart: Thank you for working on this

pm5k:
> @pm5k Just wanted to drop in and say, amazing work. It looks splendid ❤️ Thank you for working on this

Thank you. I sort of left it here until I get a green light from someone. There’s a few changes to be made, but ultimately it would be nice to know whether there is a chance for this to get merged so I can just wrap this up. Of course it would also be good to get some critique from the author of the project as ultimately this would have to fit with their vision for pdoc. Glad you liked it though. :)

chwba:
If we set `hljs_style_tuple = ('atom-one-dark', 'github')`, we will get dark code highlighting and light pages - wanted to put dark mode to default and found that out.

pm5k:
> If we set `hljs_style_tuple = ('atom-one-dark', 'github')`, we will get dark code highlighting and light pages - wanted to put dark mode to default and found that out.

That’s purely because the first item is light and the second is dark. In order to set dark mode as default you would have to either swap the vars in the theme or I would have to expose that as an option. Right now it works in such a way that dark mode can only be set as default by toggling it once in the doc switch once rendered out. Good catch though. I will add it in at some point.

PR #354, opened 2021-08-22
BUG: Fix crash on `NotImplementedError`

### LINKED ISSUE #353: Pdoc fails if a base class uses a NotImplementedError
### Expected Behavior

No error thrown. I have a base class in a lib that I want to document and some of the methods are deliberately marked as `NotImplementedError`.

### Actual Behavior

`_getmembers_all` catches only `AttributeError`. It should also catch `NotImplementedError`

### Steps to Reproduce

1. Take any class in an existing library and insert a `raise` statement for a `NotImplementedError`
2. Run `pdoc --html some_lib` where `some_lib` is the name of the aforementioned lib.

### Additional info

- pdoc version: 0.10.0

kjczarne:
Closes #353 

It's a simple fix. Didn't seem to affect the unit tests run. Lmk if this is ok. 😄

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -987,7 +987,7 @@ def _getmembers_all(obj: type) -> List[Tuple[str, Any]]:
     for name in names:
         try:
             value = getattr(obj, name)
-        except AttributeError:
+        except (AttributeError, NotImplementedError):
             for base in mro:
                 if name in base.__dict__:
                     value = base.__dict__[name]


kernc:
> Didn't seem to affect the unit tests run.

Fixing the bug didn't affect the current unit tests because there are no tests in place preventing the introduction of the same bug in the future. If you'd care to add a simple test showing that documenting a class method that raises NotImplementedError now works (and previously didn't), I'd in fact love to see this through. :sweat_smile:

PR #367, opened 2021-11-08
Update README.md

TarunavBA:
Updated the usage and installation steps


TarunavBA:
Same author as #366

kernc:
The author is indeed the same. Unfortunately, so is the quality of the contribution.

PR #385, opened 2022-02-01
Leave trailing line break in reST directives

### LINKED ISSUE #384: Tables in `.. include::`ed markdown files merge with remaining docstring
### Expected Behavior

When pulling in a markdown table via [`.. include::` reStructuredText directive](https://pdoc3.github.io/pdoc/doc/pdoc/#supported-rest-directives), any remaining content within that docstring simply follows on _after_ said table.
expected

### Actual Behavior

The remaining content gets included within the table from the `.. include::`ed markdown table.

actual

### Steps to Reproduce

1. Extract the contents of  [pdoc3-test.zip](https://github.com/pdoc3/pdoc/files/7970493/pdoc3-test.zip) and open a terminal within the top level directory.
2. Install `pdoc3` (I discovered this whilst using 0.9.2, but the issue is also present in 0.10.0
3. Run `pdoc --html --output-dir doc package --force`
4. Observe `doc/package/index.html. You'll notice that everything in the top-level docstring from `package/__init__.py` after the `.. include::`ed markdown table from `package/table.md` is included in the rendered table, as opposed to _after_ the table, as expected.

### Additional info

- pdoc version: 0.9.2 and/ or 0.10.0
- Workaround: add an empty HTML tag of some form (tested with `span`) after the table.

howamith:
Closes #384.

Feels a little clunky adding a trailing line break when processing reST directives, to then have to remove it again for reST directives in code blocks, but I couldn't see any other way. Figured I'd submit a PR anyway for feedback and go from there.

### DIFF
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -326,7 +326,7 @@ class _ToMarkdown:
         See: https://python-markdown.github.io/extensions/admonition/
         """
         substitute = partial(re.compile(r'^(?P<indent> *)\.\. ?(\w+)::(?: *(.*))?'
-                                        r'((?:\n(?:(?P=indent) +.*| *$))*[^\n|\r|\r\n])*',
+                                        r'((?:\n(?:(?P=indent) +.*| *$))*[^\r\n])*',
                                         re.MULTILINE).sub,
                              partial(_ToMarkdown._admonition, module=module,
                                      limit_types=limit_types))
--- a/pdoc/html_helpers.py
+++ b/pdoc/html_helpers.py
@@ -98,6 +98,11 @@ def _fenced_code_blocks_hidden(text):
     def hide(text):
         def replace(match):
             orig = match.group()
+
+            # Remove a trailing blank line before the closing ```, which might
+            # have been added while processing `.. include::` directives.
+            orig = orig.replace("\n\n```", "\n```")
+
             new = f'@{hash(orig)}@'
             hidden[new] = orig
             return new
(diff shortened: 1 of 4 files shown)


pdoc/html_helpers.py
```diff
@@ -98,6 +98,11 @@ def _fenced_code_blocks_hidden(text):
     def hide(text):
         def replace(match):
             orig = match.group()
+
+            # Remove a trailing blank line before the closing ```, which might
+            # have been added while processing `.. include::` directives.
```
kernc, line 103:
```suggestion
            # Remove a trailing blank line before the closing ```, which was
            # added if the code block came in via `.. include::` directive.
```

kernc (approved):
Clunky indeed, but unless we have a way to fix the regex:
https://github.com/pdoc3/pdoc/blob/4aa70de2221a34a3003a7e5f52a9b91965f0e359/pdoc/html_helpers.py#L325-L328
I don't have a really better idea.

howamith:
Another possibly way of neatening this up would be to somehow detect the context of where the reST directive is (i.e whether or not it's within a code block) and only add the trailing new-line where appropriate. This context could be detected in, or outside of and passed in, to `_ToMarkdown._admonition()`.

kernc:
Another way would be to actually amend the culprit regex. It's posed as:
```regexp
^(?P<indent> *)\.\. ?(\w+)::(?: *(.*))?
((?:\n(?:(?P=indent) +.*| *$))*)
```
for its second line to matche any indented paragraphs (separated with a line break). E.g.:
```rst
Outside admonition.

.. warning:: Title
   First paragraph.

   Second paragraph.  # Our regex matches this final line break. But if it didn't ...

Outside admonition.
```
As can be seen [here](https://regex101.com/r/O9Wvpv/1) (trailing `\n` is visible in the "Export Matches" JSON).

I think we could rewrite the regex to not consume the line break unless another indented paragraph follows it ... :thinking:

pdoc/html_helpers.py
```diff
@@ -323,7 +326,8 @@ def admonitions(text, module, limit_types=None):
         See: https://python-markdown.github.io/extensions/admonition/
         """
         substitute = partial(re.compile(r'^(?P<indent> *)\.\. ?(\w+)::(?: *(.*))?'
-                                        r'((?:\n(?:(?P=indent) +.*| *$))*)', re.MULTILINE).sub,
+                                        r'((?:\n(?:(?P=indent) +.*| *$))*[^\n|\r|\r\n])*',
```
kernc, line 329: `[]` itself denotes a disjunctive list of characters, so `[^\r\n]` suffices.
```suggestion
                                        r'((?:\n(?:(?P=indent) +.*| *$))*[^\r\n])*',
```

howamith: Told ya I wasn't a regex expert ;)

kernc:
Excellent find of the regex! Many thanks! :cake:

PR #390, opened 2022-03-29
added support for nested classes; this is basically a newer version o…

frank101010:
Adds support for nested classes to pdoc3.

This is basically a newer version of angeloskath's commit 467350c12894e66d45151e274db79e2cb7df5906,  https://github.com/pdoc3/pdoc/pull/176, but since I need support for nested classes, and the original merge request hasn't changed status since June 2020, I decided to apply most of angeloskath's changes into a branch derived from the most recent master.

(deleted section about unit test problems, which were resolved after running tests on an os supporting symbolic links)

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1020,7 +1020,8 @@ class Class(Doc):
     """
     __slots__ = ('doc', 'cls', '_super_members')
 
-    def __init__(self, name: str, module: Module, obj, *, docstring: str = None, cls: 'Class' = None):
+    def __init__(self, name: str, module: Module, obj, *, docstring: str = None,
+                 cls: 'Class' = None):
         assert inspect.isclass(obj)
 
         if docstring is None:
@@ -1030,12 +1031,12 @@ class Class(Doc):
             docstring = f'{inspect.getdoc(obj) or ""}\n\n{init_doc}'.strip()
 
         super().__init__(name, module, obj, docstring=docstring)
-        
+
         self.cls = cls
         """
         The `pdoc.Class` object if this class is defined in a class. If not,
         this is None.
-        """        
+        """
 
         self.doc: Dict[str, Union[Function, Variable, Class]] = {}
         """A mapping from identifier name to a `pdoc.Doc` objects."""
@@ -1184,7 +1185,7 @@ class Class(Doc):
         result = [obj for obj in _filter_type(type, self.doc)
                   if (include_inherited or not obj.inherits) and filter_func(obj)]
         return sorted(result) if sort else result
-    
+
     def classes(self, include_inherited=False, sort=False):
         """Returns the classes nested in this class."""
         return self._filter_doc_objs(
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -951,23 +951,12 @@ class Module(Doc):
         """
         return self._filter_doc_objs(Variable, sort)
 
-    def _classes(self, sort=True) -> List['Class']:
+    def classes(self, sort=True) -> List['Class']:
         """
-        Returns all documented module-level classes in the module,
+        Returns all documented classes in the module,
         optionally sorted alphabetically, as a list of `pdoc.Class`.
         """
         return self._filter_doc_objs(Class, sort)
-
-    def classes(self, sort=True) -> List['Class']:
-        """
-        Returns all documented classes in the module and all their
-        nested classes.
-        """
-        results = []
-        for c in self._classes(sort)[::-1]:
-            results.append(c)
-            results += c.all_nested_classes()
-        return results
 
     def functions(self, sort=True) -> List['Function']:
         """
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -712,7 +712,15 @@ class Module(Doc):
             if _is_function(obj):
                 self.doc[name] = Function(name, self, obj)
             elif inspect.isclass(obj):
-                self.doc[name] = Class(name, self, obj)
+                cl = Class(name, self, obj)
+                self.doc[name] = cl
+                # Also add all nested classes of the class just found to the
+                # module context, otherwise some classes will be recognized
+                # as "External" even though they were correctly recognized
+                # as "Class" during an earlier scanning process
+                # (=> Module.find_ident()).
+                for ncl in cl.all_nested_classes():
+                    self.doc[f'{name}.{ncl.name}'] = ncl
             elif name in var_docstrings:
                 self.doc[name] = Variable(name, self, var_docstrings[name], obj=obj)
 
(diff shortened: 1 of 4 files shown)


kernc (changes requested):
Thanks for bringing this up again!

Would you care to add a small unit test to show this new feature works as advertised and so we don't break it later?

pdoc/__init__.py
```diff
@@ -1018,7 +1032,13 @@ def __init__(self, name: str, module: Module, obj, *, docstring: str = None):
 
         super().__init__(name, module, obj, docstring=docstring)
 
-        self.doc: Dict[str, Union[Function, Variable]] = {}
+        self.cls = cls
```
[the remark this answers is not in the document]
frank101010: `Class.cls` is used in `Class.inherited_members`, the `i.cls` in the lambda expression. `Class.doc.values()` may now also contain `Class` objects in addition to `Function` and `Variable` objects.

pdoc/__init__.py
```diff
@@ -1063,6 +1083,14 @@ def definition_order_index(
             if _is_function(obj):
                 self.doc[name] = Function(
                     name, self.module, obj, cls=self)
+            elif inspect.isclass(obj):
+                self.doc[name] = Class(
+                    self.name + "." + name,
+                    self.module,
+                    obj,
+                    cls=self,
+                    docstring=inspect.getdoc(obj)
```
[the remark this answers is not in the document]
frank101010: Will be removed.

pdoc/__init__.py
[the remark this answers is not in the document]
kernc: So with:
```py
class C:
    class D:
       ...
class A(C):
    ...
```
`A.D` will show as inherited from `C`? Fine.

Do `C.D` and `A.D` both get listed on the module level? I guess that's fine, too.

pdoc/__init__.py
```diff
@@ -950,6 +950,19 @@ def classes(self, sort=True) -> List['Class']:
         """
         return self._filter_doc_objs(Class, sort)
 
+    def all_classes(self, sort=True) -> List['Class']:
```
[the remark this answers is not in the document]
frank101010: I renamed `classes()` to `_classes()`, then `all_classes()` to `classes()`, so the template was reverted to its original contents.
I also added a unit test case, `ApiTest.test_nested_classes()`.

pdoc/__init__.py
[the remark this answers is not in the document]
frank101010: Yes, all the nested classes also get documented on the module level.

frank101010:
@kernc: you might want to wait with the next review, I've discovered a problem regarding warnings being issued, which I've already resolved. But there are also cases where member types are not properly rendered as hyperlinks in html output, where I'm still looking for the cause for.

frank101010:
@kernc: I just pushed new changes which fix a problem nested classes being recognized as `pdoc.External` instead of `pdoc.Class`. This required adding the nested classes into `pdoc.Module`'s context. Unfortunately, this also breaks one unit test, namely `ApiTest.test__all__`. Since I don't really understand how that test works, could you please have a look at it?

pdoc/test/__init__.py
```diff
@@ -1026,6 +1027,41 @@ class G(F[int]):
...
+        m_module = importlib.util.module_from_spec(m_spec)
+        code = '''
+class A:
+    """Class A documentation"""
+    class B:
+        """ Class A.B documentation"""
+        class C:
+            """ Class A.B.C documentation"""
+            pass
+'''
```
kernc, line 1042: We should add another `class D(A): pass` here, so the inheritance behavior is also tested.

pdoc/test/__init__.py
```diff
@@ -1026,6 +1027,41 @@ class G(F[int]):
...
+
+        A = m_module.A
+        B = A.B
+        C = B.C
+
+        sys.modules[m_name] = m_module
+
+        try:
+            mod = pdoc.Module('M', context=pdoc.Context())
+            print([c.qualname for c in mod.classes()])
```
kernc, line 1053:
```suggestion
```

pdoc/__init__.py
```diff
@@ -1195,6 +1225,19 @@ def functions(self, include_inherited=True, sort=True) -> List['Function']:
             Function, include_inherited, lambda dobj: not dobj.is_method,
             sort)
 
+    def all_nested_classes(self, include_inherited=True, sort=True) -> List['Class']:
```
kernc, line 1228: With friendly `Class.classes()` available, let's make this method a private implementation detail?
```suggestion
    def _nested_classes(self, include_inherited=True, sort=True) -> List['Class']:
```

pdoc/__init__.py
```diff
@@ -712,7 +712,15 @@ def is_from_this_module(obj):
...
-                self.doc[name] = Class(name, self, obj)
+                cl = Class(name, self, obj)
+                self.doc[name] = cl
+                # Also add all nested classes of the class just found to the
+                # module context, otherwise some classes will be recognized
+                # as "External" even though they were correctly recognized
+                # as "Class" during an earlier scanning process
+                # (=> Module.find_ident()).
+                for ncl in cl.all_nested_classes():
+                    self.doc[f'{name}.{ncl.name}'] = ncl
```
kernc, line 723: Does using `f'{ncl.refname}'` work?

kernc:
The test tries to ensure that only the identifiers in [`__all__`](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) are exposed. Interestingly, I'm not sure where the value [`'B.B.C'`](https://github.com/pdoc3/pdoc/runs/5848844773?check_suite_focus=true#step:10:120) comes from. I'd understand `'B.C'` since that's what's nested. :thinking: 
https://github.com/pdoc3/pdoc/blob/2cce30a9b55eeeddc1ed826c8a2ada53777c3eea/pdoc/test/example_pkg/__init__.py#L111-L112

pdoc/__init__.py
frank101010: Done.

pdoc/__init__.py
frank101010: Yes, and that actually solves `test__all__()`, which now reports `B` and `B.C`.

pdoc/__init__.py
```diff
@@ -1158,6 +1180,14 @@ def _filter_doc_objs(self, type: Type[T], include_inherited=True,
                   if (include_inherited or not obj.inherits) and filter_func(obj)]
         return sorted(result) if sort else result
 
+    def classes(self, include_inherited=False, sort=False):
```
kernc, line 1183: With other methods, `include_inherited=` (and `sort=`) default is True. I noticed because `D.B` and `D.B.C` don't get listed in the unit test.

frank101010: Ok, I had to debug this. Now I know that I have to call `pdoc.link_inheritance()` before I can query `Class` objects for their inherited nested classes (unless I'm doing that, `Class.classes(D)` will not report the nested classes inherited from `A`.

I have found out the following:

* `Module.classes()` reports all classes in the module, even the nested ones.
* `Class.classes()` just reports only the immediate nested classes (with `A`, `A.B`, `A.B.C`, `A.classes()` reports just `[A.B]` , not `[A.B, A.B.C]`)

Is this behavior acceptable or should we adapt the methods' names appropriately?

Issue #392, opened 2022-04-04
pdoc3 uses the soon to be deprecated distutils package


991jo:
Apparently distutils is getting deprecated in Python 3.12. For more details see [PEP-632](https://peps.python.org/pep-0632/)

### Expected Behavior

Not getting spammed with deprecation warnings.

### Actual Behavior

I see some deprecation warnings.
```
$ pdoc3 <my module here>
<path to venv>/lib/python3.10/site-packages/pdoc/cli.py:490: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.sysconfig import get_python_lib
<path to venv>/lib/python3.10/site-packages/pdoc/cli.py:490: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
  from distutils.sysconfig import get_python_lib
<actual command output>
```

### Steps to Reproduce

1. Execute pdoc3 <module> with python3.10 on a module. Simply running the `--help` or `--version` commands does not work to trigger the deprecation warnings.

### Additional info

- pdoc version: `pdoc3 0.10.0`
- python version: 3.10.2

kernc:
Thanks. I guess the needed work entails replacing uses of `distutils.sysconfig.get_python_lib()`:
https://github.com/pdoc3/pdoc/blob/f358893e4fcfd7f29857a7ff5491b606ff146d39/pdoc/cli.py#L490-L491
https://github.com/pdoc3/pdoc/blob/2cce30a9b55eeeddc1ed826c8a2ada53777c3eea/pdoc/html_helpers.py#L625-L627
with appropriate calls to [`sysconfig.get_path()`](https://docs.python.org/3/library/sysconfig.html#sysconfig.get_path)?

braniii:
I guess this is the relevant replacement function `sysconfig.get_path('platlib')`, see https://bugs.python.org/issue41282#msg393018 for further infromations.

timkofu:
PR https://github.com/pdoc3/pdoc/pull/401

LostInDarkMath:
Is there any progress here? It seems like `pdoc3` does not work with Python 3.12 due to this issue: `pdoc3 --html --output-dir docs pedantic --force` gives the error
```
Traceback (most recent call last):
  File "/home/willi/Documents/GitHub/pedantic-python-decorators/.venv/bin/pdoc3", line 8, in <module>
    sys.exit(main())
[...]

PR #401, opened 2022-05-16
Replaced distutils.sysconfig.get_python_lib with sysconfig.get_path

timkofu:
Fixes https://github.com/pdoc3/pdoc/issues/392

As `distutils.sysconfig` is now deprecated.

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -487,8 +487,8 @@ def main(_args=None):
         pass  # pdoc was not invoked while in a virtual environment
     else:
         from glob import glob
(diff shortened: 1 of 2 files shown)


timkofu:
Won't work with 3.6.

kernc:
Thanks for the PR. I don't think that's necessarily why the CI is failing. Anyway, [Python 3.6 EOL'd a couple of months ago](https://endoflife.date/python), so that's not an issue.

timkofu:
@kernc Needed a bit more work; https://github.com/pdoc3/pdoc/pull/402

PR #408, opened 2022-08-13
Skip `__editable__` paths during `iter_modules`

abhinavsingh:
Document generation seems to be broken for `namespace` packages after recent changes to how `__editable__` install work.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -727,6 +727,8 @@ class Module(Doc):
                 """
                 from os.path import isdir, join
                 for pth in paths:
+                    if pth.startswith("__editable__"):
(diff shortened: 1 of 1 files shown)


pdoc/__init__.py
```diff
@@ -727,6 +727,8 @@ def iter_modules(paths):
                 """
                 from os.path import isdir, join
                 for pth in paths:
+                    if pth.startswith("__editable__"):
```
kernc, line 730:
```suggestion
                    if pth.startswith("__editable__."):
                        # See https://github.com/pypa/pip/issues/11380
```

kernc (approved):

Issue #310, opened 2021-02-13
Use of Nazi symbols in the footer on the project website


Operator2024:
Hi people! I found Nazi symbols https://pdoc3.github.io/pdoc/ in the footer. Do you think this is normal? or is it just a typo
There's a screenshot - https://1drv.ms/u/s!AtV6DG59xKaXrlO49xdqCON6JGpC?e=KJv1FI

kernc:
They're _not_ Nazi symbols, mind you. Rather, they are [_Buddhist_ symbols](https://en.wikipedia.org/wiki/Swastika#Buddhism). :sweat_smile: It's fairly easy to mistake the two, but I don't think the Unicode Consortium would ever agree to add [symbol characters](https://en.wikipedia.org/wiki/Swastika#Written_characters) that glorify Nazism.

Operator2024:
Ok, sorry guys! I didn't think about the Unicode consortium ... maybe you're right

cgobat:
The unfortunate fact of the matter is that when people see that symbol on the project's webpage, Buddhism is *not* going to be the first thing that comes to mind. The very same Wikipedia page you linked states

> it continues to be strongly associated with [Nazism](https://en.wikipedia.org/wiki/Nazism), [antisemitism](https://en.wikipedia.org/wiki/Antisemitism), [white supremacism](https://en.wikipedia.org/wiki/White_supremacy), or simply [evil](https://en.wikipedia.org/wiki/Evil)

I cannot imagine that this is a desirable impression to be giving to site visitors, especially given the lack of context or logical reason for why the symbol is there. @kernc, could you take a look at #424?

PR #424, opened 2023-06-14
Replace '卐' in `credits.mako` with actual credits

cgobat:
Regardless of the intent and the explanations that @kernc provided in #152 and #310, the unfortunate reality (in much of the world) is that swastikas today are much more widely recognized as/associated with Nazism than with Buddhism or anything benign. Note the following line from [Wikipedia](https://en.wikipedia.org/wiki/Swastika):

> in the West it continues to be strongly associated with [Nazism](https://en.wikipedia.org/wiki/Nazism), [antisemitism](https://en.wikipedia.org/wiki/Antisemitism), [white supremacism](https://en.wikipedia.org/wiki/White_supremacy), or simply [evil](https://en.wikipedia.org/wiki/Evil).

Given the lack of context/explanation regarding its presence on the website, it comes across at *best* as being in poor taste and at worst as a covert endorsement of Nazi ideology. As far as I can tell, there is no good reason for it to be there, so I think it should be removed. This PR replaces it with content that utilizes the `credits.mako` file/template as intended.

### DIFF
--- a/doc/pdoc_template/credits.mako
+++ b/doc/pdoc_template/credits.mako
@@ -1,1 +1,1 @@
-<p><span style="color:#ddd">&#21328;</span></p>
+<p><span style="color:#ddd"><a href="https://github.com/kernc">@kernc</a> and <a href="https://github.com/pdoc3/pdoc/graphs/contributors">others</a></span></p>


cgobat:
@kernc bump

cgobat:
It looks like this has [not actually been completed](https://github.com/pdoc3/pdoc/blob/master/doc/pdoc_template/credits.mako). @kernc, was this closed on accident?

cgobat:
It looks like this has [not actually been completed](https://github.com/pdoc3/pdoc/blob/master/doc/pdoc_template/credits.mako). @kernc, was this closed on accident?

PR #430, opened 2023-11-26
feat: separate class and static methods in html.mako

### LINKED ISSUE #126: classmethods documented as a static method
### Expected Behavior

image

### Actual Behavior

image

### Steps to Reproduce

Example code:

```
import documentation_test.submod1.submod1 as submod1
import documentation_test.submod2.submod2 as submod2

__all__ = ['testFunction', 'testClass']


class testClass(object):

    """Summary of class

    Attributes
    ----------
    arg : str
        a string to print
    """

    def __init__(self, arg: str):
        """Summary of init

        Parameters
        ----------
        arg : str, optional
            a string to print
        """
        super(testClass, self).__init__()
        self.arg = arg

    @classmethod
    def createString(cls, aString: str):
        """create a string

        Parameters
        ----------
        aString : str
            a string to print
        """
        return cls(arg=aString)

    def printString(self):
        """Print a string
        """
        print(self.arg)


def testFunction(anything: str = 'test') -> None:
    """test function for a demo import of whole module

    Parameters
    ----------
    anything : str, optional
        a test string

    Returns
    -------
    None
    """
    print(anything)


submod1.submod1('string from submod1 in base')

submod2.submod2('string from submod2 in base')
```

Command ran: `pdoc --config show_type_annotations=True --html src/documentation_test --force`

### Additional info

- pdoc version: 0.7.2

Not sure if this is intentional or I have done something nonstandard. No matter what I try; class in a submodule, class with both static and class methods, etc. I get the same result.

JPHutchins:
Fixes #126 

I'm afraid I don't know much about Mako, so I didn't add class method sections PDF or text.

Future discussion: IMO, the ordering of method sections should go from most-specific to least-specific:

- Instance methods
- Class methods
- Static methods

There is one failing test on my local but it seems unrelated to this change, please LMK!

Cheers,
JP

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1186,17 +1186,32 @@ class Class(Doc):
         represent this class' methods.
         """
         return self._filter_doc_objs(
-            Function, include_inherited, lambda dobj: dobj.is_method,
+            Function, include_inherited, lambda dobj: dobj.is_instance_method,
             sort)
 
-    def functions(self, include_inherited=True, sort=True) -> List['Function']:
+    def static_methods(self, include_inherited=True, sort=True) -> List['Function']:
         """
         Returns an optionally-sorted list of `pdoc.Function` objects that
-        represent this class' static functions.
+        represent this class' static methods.
         """
         return self._filter_doc_objs(
-            Function, include_inherited, lambda dobj: not dobj.is_method,
-            sort)
+            Function,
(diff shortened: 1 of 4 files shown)


JPHutchins:
https://github.com/mitmproxy/pdoc#pdoc-vs-pdoc3

PR #431, opened 2023-11-26
Feature/better enums

JPHutchins:
Generally, `Enums` in python are used a bit differently than classes and IMO should be treated distinctly.  Primarily, the "variables" should display what the enumeration corresponds to and they should not be sorted by default.

Here's a screenshot of how this PR formats a library I am working on:

image

This PR implements simple parsing for Enums.  Initially, I subclassed `Class`.  It ended up not really being necessary within the scope of this PR, but nevertheless could provide a nice place to extend functionality if necessary.


JPHutchins:
https://github.com/mitmproxy/pdoc#pdoc-vs-pdoc3

Issue #436, opened 2024-01-04
Deprecation warning for PEP224 class variable docstrings


heseber:
### Expected Behavior
Create output for Python files with PEP224 class variable docstrings without warnings.

### Actual Behavior

- .../pdoc/__init__.py:260: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  isinstance(str_node.value, ast.Str)):
- .../pdoc/__init__.py:286: DeprecationWarning: Attribute s is deprecated and will be removed in Python 3.14; use value instead
  docstring = inspect.cleandoc(str_node.value.s).strip()

### Steps to Reproduce

1. Create a python file with a class with a class variable and annotate the class variable with a PEP224 string
2. Run pdoc3 on that file 

### Additional info

- pdoc version: 0.8.1

PR #437, opened 2024-01-04
Fix deprecation warnings for PEP224 docstrings of class variables

heseber:
Fix for https://github.com/pdoc3/pdoc/issues/436


kernc:
Thanks!

PR #438, opened 2024-02-13
BUG: Handle import of distutils for python 3.12

ERosendo:
This PR adds code to handle the import of the `sysconfig` module for Python >= 3.11

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -487,8 +487,12 @@ def main(_args=None):
         pass  # pdoc was not invoked while in a virtual environment
     else:
         from glob import glob
-        from distutils.sysconfig import get_python_lib
-        libdir = get_python_lib(prefix=venv_dir)
+        if sys.version_info >= (3, 11):
+            from sysconfig import get_path
+            libdir = get_path("platlib")
+        else:
+            from distutils.sysconfig import get_python_lib
+            libdir = get_python_lib(prefix=venv_dir)
         sys.path.append(libdir)
         # Resolve egg-links from `setup.py develop` or `pip install -e`
         # XXX: Welcome a more canonical approach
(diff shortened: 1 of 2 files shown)


kernc (changes requested):

ferdnyc:
IIUC correctly, based on the comment above the code that uses `distutils`, it's there to support one very particular use case: When the user is working in a virtualenv, but runs a `pdoc` script installed _outside_ of that virtualenv (in a system or user path).

Honestly, the most sensible solution, to me, seems to be: **Just don't support that.** It's a lot of effort for very little reward. When working in a venv, install `pdoc3` in the venv. Simple as that, and all of the path shenanigans go away.

_Otherwise_, [checking for a `VIRTUAL_ENV` environment variable is the wrong way to detect when running in a venv](https://stackoverflow.com/q/15466253/200794) (though it is a clever way to detect a non-venv-installed package being run from an activated venv).

But if `pdoc` is being executed under a _non_-venv Python interpreter, even with an activated venv, `distutils.sysconfig.get_python_lib()` / `sysconfig.get_path()` won't return the paths inside the venv. They'll return the system paths.

And if `pdoc` is actually being executed in a venv (that it's installed in), then there's no need to screw with `sys.path`, it'll already point into the venv's `site-packages` dir. If there are any editable installs, `importlib.import_module('packagename')` will work to import those packages, and the path to the source files is just `os.path.dirname(packagename.__path__[0])`.

So I really don't get what that code is trying to do anyway.

kernc:
> And if `pdoc` is actually being executed in a venv (that it's installed in), then there's no need to screw with `sys.path`, it'll already point into the venv's `site-packages` dir. If there are any editable installs, `importlib.import_module('packagename')` will work to import those packages, and the path to the source files is just `packagename.__path__[0]`.

@ferdnyc Apparenly, this being a non-issue hasn't always been the case ... 🤷

Issue #441, opened 2024-06-21
"x | y" does not work but "Union[x, y]" does


AnthonyZJiang:
e.g. for a method:

```
def get_val() -> int | str:
    """Get value.

    Returns:
        int | str: the value.
    """
    pass
```

### Expected Behavior

def get_val() ‑> int | str
Get value.

**Returns**
    _int | str_
The value.

### Actual Behavior

def get_val() ‑> int | str
Get value.

**Returns** 
    int | None The value.

See more details in the attached picture

### Additional info

- pdoc version: 0.10.0

# Example 1
Here is the example in action, comparing | and Union:
![image](https://github.com/pdoc3/pdoc/assets/8778250/e6aed3fd-61f1-4fe9-8e06-068130a510f9)

# Example 2
Here's another example of using | instead of Union:
![image](https://github.com/pdoc3/pdoc/assets/8778250/bc27a5e6-911d-41ea-8fba-badbd6d9c164)

ERosendo:
@kernc is Anyone working on a fix for this yet? I'd be happy to take a look and see if I can help fix it.

ERosendo:
@AnthonyZJiang, can you rerun the command and let us know if the problem you encountered earlier is fixed? Any additional details you can give would be helpful if it's still happening.

AnthonyZJiang:
The issue is solved. The doc was generated correctly. I appreciate your dedication and fast response.

PR #443, opened 2024-06-22
ENH: Handle union type expressions for Google style docstrings

ERosendo:
This PR addresses [issue #441](https://github.com/pdoc3/pdoc/issues/441) by tweaking the regex expression used to parse Google-style docstrings to handle union-type expressions correctly. 

I also updated the test for this docstring format to ensure the fix works as expected and prevent regressions in the future.


kernc:
Looks good, thanks!

Issue #450, opened 2024-08-19
setting `git_link_template` causes failures on `property` and `cached_property`


mivanit:
### Expected Behavior

pdoc should successfully build docs with `git_link_template = 'https://github.com/USER/PROJECT/blob/{commit}/{path}#L{start_line}-L{end_line}'` for classes with `property` or `cached_property`

### Actual Behavior

```python
Python/Lib/site-packages/pdoc/html_helpers.py:580: UserWarning: format_git_link for <functools.cached_property object at ...> failed:
Traceback (most recent call last):
  File "Python/Lib/site-packages/pdoc/html_helpers.py", line 567, in format_git_link
    abs_path = inspect.getfile(inspect.unwrap(dobj.obj))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Python/Lib/inspect.py", line 929, in getfile
    raise TypeError('module, class, method, function, traceback, frame, or '
TypeError: module, class, method, function, traceback, frame, or code object was expected, got cached_property

  warn(f'format_git_link for {dobj.obj} failed:\n{traceback.format_exc()}')
```

### Steps to Reproduce

1. create a python module with a class with a `property` or `cached_property`
2. create a `config.mako` with `git_link_template` not `None`
3. run pdoc on the module with the `config.mako` in the templates dir

### Additional info

- pdoc version: `0.11.1`

Lucas-C:
I confirm that I also see this warning when using `pdoc` to render `fpdf2` documentation: https://github.com/py-pdf/fpdf2/actions/runs/12265238360/job/34220855578

PR #451, opened 2024-08-19
fix properties and cached_properties failing when adding git links

mivanit:
Fixes https://github.com/pdoc3/pdoc/issues/450


kernc:
I did a little moving around, but thanks for figuring out the hard stuff!

PR #455, opened 2025-01-05
Fix broken anchor tags in --pdf mode

### LINKED ISSUE #389: anchor tags in markdown output are not being resolved
I'm not sure whether this is a bug or simply I'm missing some configuration option, but if its the latter I can't find it so I apologise. I'm using the `--pdf` option to get a markdown document, but every section in it seems to have the same anchor tag, namely 'id' , for example one of the headers in the produced file is:

##### Method `fit_tail_model` {#id}

where the tag should be something like `package.module.class.fit_tail_model` instead of `id` .

### Expected Behavior
producing a markdown file where anchor tags are resolved so that reference within the same document work

### Actual Behavior
All tags seem to be set to a default string of 'id', so module/class/method links don't work. Strangely, for html outputs this  works fine.

### Steps to Reproduce

1. `pdoc3 --pdf -c latex_math=True mymodule > docs.md`

### Additional info

- pdoc version: 0.10.0

Patrik112358:
original fix suggested by @LauraLMann in https://github.com/pdoc3/pdoc/issues/389#issuecomment-1108855774

Fixes #389

### DIFF
--- a/pdoc/templates/pdf.mako
+++ b/pdoc/templates/pdf.mako
@@ -27,7 +27,7 @@
 %>
 
 <%def name="title(level, string, id=None)">
-    <% id = f' {{#id}}' if id is not None else '' %>
+    <% id = f' {{#{id}}}' if id is not None else '' %>
 ${('#' * level) + ' ' + string + id}
 </%def>
 


kernc:
Thanks for bringing it up!

PR #464, opened 2025-07-24
BUG: Rewrite `inspect.unwrap()` to respect classes

### LINKED ISSUE #463: Ambigous usage of `inspect.unwrap()` on classes
https://github.com/pdoc3/pdoc/blob/f8704ec70c54d6d9fe358f8fabdb49eb6fd63580/pdoc/__init__.py#L733

PDoc uses `inspect.unwrap()` in multiple places. However, this function is designed to work on functions and is not able to work on classes properly.

### 3.12.1
Before 3.12.3, `inspect.unwrap()` would go on a class as long as `__wrapper__` field exists, which is fine for functions, but for classes, it goes up to the parent class if you inherited from a decorated class. This results in PDoc rendering wrong docstrings for classes.

Class defition

```
PyDev console: starting.
Python 3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)] on win32
from pdoc_issues.class_wrapper import AlwaysOne
AlwaysOne().value
1
AlwaysOne.__doc__
" This is an `AbstractValue`'s implementation that always returns 1. "
```

However, ...
```
D:\Projects\test\.venv.d\venv-3.12.1\Scripts\python.exe -m pdoc pdoc_issues.class_wrapper --html -o html 
html\pdoc_issues\class_wrapper\index.html
html\pdoc_issues\class_wrapper\class_definition.html
html\pdoc_issues\class_wrapper\util.html

Process finished with exit code 0
```

Image

Investigation/debugging result: `obj` is evaluated to be the parent class instead of the actual child, and thus stored in context on a child's name.

### 3.12.10
Starting from 3.12.3, this was changed. From now on, `inspect.unwrap()` just does not unwrap a class. Which is crucial for the PDoc's ability to extract docstrings and other metadata.

Image

https://docs.python.org/release/3.12.3/whatsnew/changelog.html#library
https://github.com/python/cpython/issues/112006

Attempting to generate documentation just fails.

```
D:\Projects\test\.venv.d\venv-3.12.10\Scripts\python.exe -m pdoc pdoc_issues.class_wrapper --html -o html 
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "D:\Projects\test\.venv.d\venv-3.12.10\Lib\site-packages\pdoc\__main__.py", line 6, in <module>
    main()
  File "D:\Projects\test\.venv.d\venv-3.12.10\Lib\site-packages\pdoc\cli.py", line 559, in main
    pdoc.link_inheritance()
  File "D:\Projects\test\.venv.d\venv-3.12.10\Lib\site-packages\pdoc\__init__.py", line 498, in link_inheritance
    for cls in _toposort(graph):
               ^^^^^^^^^^^^^^^^
  File "D:\Projects\test\.venv.d\venv-3.12.10\Lib\site-packages\pdoc\__init__.py", line 480, in _toposort
    assert not graph, f"A cyclic dependency exists amongst {graph!r}"
           ^^^^^^^^^
AssertionError: A cyclic dependency exists amongst {<Class 'pdoc_issues.class_wrapper.class_definition.AbstractValue'>: {<Class 'pdoc_issues.class_wrapper.class_definition.AbstractValue'>}, <Class 'pdoc_issues.class_wrapper.class_definition.AlwaysOne'>: {<Class 'pdoc_issues.class_wrapper.class_definition.AlwaysOne'>, <Class 'pdoc_issues.class_wrapper.class_definition.AbstractValue'>}, <Class 'pdoc_issues.class_wrapper.AbstractValue'>: {<Class 'pdoc_issues.class_wrapper.class_definition.AbstractValue'>}, <Class 'pdoc_issues.class_wrapper.AlwaysOne'>: {<Class 'pdoc_issues.class_wrapper.class_definition.AlwaysOne'>, <Class 'pdoc_issues.class_wrapper.class_definition.AbstractValue'>}}

Process finished with exit code 1
```

#### File `pdoc_issues/class_wrapper/class_definition.py`
```python
"""
This module exports the following classes:

* `AbstractValue`
* `AlwaysOne`
"""

from .util import decorate_class
from abc import ABC, abstractmethod

@decorate_class
class AbstractValue(ABC):
    """ This is `AbstractValue` class. """
    
    @abstractmethod
    def __value__(self) -> int:
        """ An `AbstractValue`'s value implementation, abstract method. """    
        raise NotImplemented
    
    @property
    def value(self) -> int:
        """ This is `AbstractValue`'s property. """
        return self.__value__()

@decorate_class
class AlwaysOne(AbstractValue):
    """ This is an `AbstractValue`'s implementation that always returns 1. """
    
    def __value__(self) -> int:
        return 1


__all__ = \
[
    'AbstractValue',
    'AlwaysOne',
]
```

#### File `pdoc_issues/class_wrapper/util.py`
```
import functools
import types
from typing import *

def wrap_first[C](cls: Type[C]) -> Type[C]:
    wrapped = types.new_class(cls.__name__, (cls, ), {})
    wrapped = functools.update_wrapper(wrapped, cls, updated=())
    wrapped = cast(Type[C], wrapped)
    
    return wrapped

def wrap_second[C](cls: Type[C]) -> Type[C]:
    wrapped = type(cls.__name__, cls.__mro__, dict(cls.__dict__))
    wrapped = functools.update_wrapper(wrapped, cls, updated=())
    wrapped = cast(Type[C], wrapped)
    
    return wrapped

def decorate_class[C](cls: Type[C]) -> Type[C]:
    """ Creates a two-step class decoration. """
    
    wrapped = wrap_first(cls)
    wrapped_again = wrap_second(wrapped)
    wrapped_again.__decorated__ = True
    
    return wrapped_again


__all__ = \
[
    'decorate_class',
]
```

### Potential Solution
Write your own `inspect.unwrap()`, which handles classes specifically:
* For functions, the behaviour is unchanged
* For classes, you should unwrap as long as `__module__` and `__name__` are preserved

### P.S.
I know generating new classes at runtime during decoration is a bad practice, but it's not my code, I just found that while using a third-party library (`dataclassabc`)

USSX-Hares:
Closes #463

### Changes
* Added `pdoc._unwrap_object()` function which behaves like `insepct.unwrap()`, but respects classes
  + This function is used instead of `inspect.unwrap()` in the entire codebase
* Added a test covering this specific case (`pdoc.test.ApiTest.test_class_wrappers` and `pdoc.test.example_pkg._test_classwrap`)

### Code Quality
* `flake8` checks passed locally (added a single exception for long lines due to longer docstrings)
  + There was a single error on `pdoc.cli#603`, but I have not touched this file
* Tests passed via `tox` on the following environments (os: Windows):
  + Python 3.9.12
  + Python 3.10.4
  + Python 3.11.9
  + Python 3.12.1
  + Python 3.12.10
  + Python 3.13.5
* There are two `mypy` issues in the code I haven't touched, however

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -449,6 +449,39 @@ def _unwrap_descriptor(dobj):
     return getattr(obj, '__get__', obj)
 
 
+def _unwrap_object(obj: T) -> T:
+    """
+    This is a modified version of `inspect.unwrap()` that properly handles classes.
+
+    Follows the chains of `__wrapped__` attributes, until either:
+    1. `obj.__wrapped__` is missing or None
+    2. `obj` is a class and `obj.__wrapped__` has a different name or module
+    """
+
+    orig = obj  # remember the original func for error reporting
+    # Memoise by id to tolerate non-hashable objects, but store objects to
+    # ensure they aren't destroyed, which would allow their IDs to be reused.
+    memo = {id(orig): orig}
+    recursion_limit = sys.getrecursionlimit()
+    while hasattr(obj, '__wrapped__'):
+        candidate = obj.__wrapped__
+        if candidate is None:
+            break
+
+        if isinstance(candidate, type) and isinstance(orig, type):
+            if not (candidate.__name__ == orig.__name__
+                    and candidate.__module__ == orig.__module__):
+                break
+
+        obj = typing.cast(T, candidate)
+        id_func = id(obj)
+        if (id_func in memo) or (len(memo) >= recursion_limit):
+            raise ValueError('wrapper loop when unwrapping {!r}'.format(orig))
+        memo[id_func] = obj
+
+    return obj
+
+
 def _filter_type(type: Type[T],
                  values: Union[Iterable['Doc'], Mapping[str, 'Doc']]) -> List[T]:
     """
@@ -712,11 +745,11 @@ class Module(Doc):
                          "exported in `__all__`")
                 else:
                     if not _is_blacklisted(name, self):
-                        obj = inspect.unwrap(obj)
+                        obj = _unwrap_object(obj)
                     public_objs.append((name, obj))
         else:
             def is_from_this_module(obj):
-                mod = inspect.getmodule(inspect.unwrap(obj))
+                mod = inspect.getmodule(_unwrap_object(obj))
                 return mod is None or mod.__name__ == self.obj.__name__
 
             for name, obj in inspect.getmembers(self.obj):
@@ -730,7 +763,7 @@ class Module(Doc):
                         self._context.blacklisted.add(f'{self.refname}.{name}')
                         continue
 
-                    obj = inspect.unwrap(obj)
+                    obj = _unwrap_object(obj)
                     public_objs.append((name, obj))
 
             index = list(self.obj.__dict__).index
@@ -1066,7 +1099,7 @@ class Class(Doc):
                     self.module._context.blacklisted.add(f'{self.refname}.{_name}')
                     continue
 
-                obj = inspect.unwrap(obj)
+                obj = _unwrap_object(obj)
                 public_objs.append((_name, obj))
 
         def definition_order_index(
@@ -1428,7 +1461,7 @@ class Function(Doc):
         try:
             # Both of these are required because coroutines aren't classified as async
             # generators and vice versa.
-            obj = inspect.unwrap(self.obj)
+            obj = _unwrap_object(self.obj)
             return (inspect.iscoroutinefunction(obj) or
                     inspect.isasyncgenfunction(obj))
         except AttributeError:
--- a/pdoc/test/__init__.py
+++ b/pdoc/test/__init__.py
@@ -659,6 +659,47 @@ class ApiTest(unittest.TestCase):
             self.assertEqual(cm, [])
             self.assertNotIn('downloaded_modules', mod.doc)
 
+    # flake8: noqa: E501 line too long
+    def test_class_wrappers(self):
+        """
+        Check that decorated classes are unwrapped properly.
+        Details: https://github.com/pdoc3/pdoc/issues/463
+        """
+
+        module_name = f'{EXAMPLE_MODULE}._test_classwrap'
+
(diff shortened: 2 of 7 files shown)


USSX-Hares:
Hi, @kernc, could you review this PR?

USSX-Hares:
Hi, @kernc, could you review this PR?

pdoc/test/__init__.py
```diff
@@ -659,6 +659,47 @@ def test__pdoc__dict(self):
             self.assertEqual(cm, [])
             self.assertNotIn('downloaded_modules', mod.doc)
 
+    # flake8: noqa: E501 line too long
```
kernc: Looking at flake8 docs, I don't see any mention of it being able to ignore errors in _blocks_.
https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors

Would you move this `noqa` to the offending lines?

pdoc/__init__.py
```diff
@@ -449,6 +449,43 @@ def _unwrap_descriptor(dobj):
...
+    """
+
+    orig = obj  # remember the original func for error reporting
+    # Memoise by id to tolerate non-hashable objects, but store objects to
+    # ensure they aren't destroyed, which would allow their IDs to be reused.
+    memo = {id(orig): orig}
+    recursion_limit = sys.getrecursionlimit()
+    while hasattr(obj, '__wrapped__'):
+        if stop is not None and stop(obj):
+            break
```
kernc: Sincere apologies for the late review! The PR generally looks good! Since this parameter is merely copied over from CPython but never used in our sources, furthermore with this function marked internal (👍), I think I would prefer to see it removed.

kernc (changes requested):

pdoc/test/__init__.py
USSX-Hares: Yeah, back in the days I was new to flake8, and this directive seems to disable all inspections in the file, we'll rewrite.

kernc:
Lgtm, thanks!

